PR #8, opened 2016-06-16
Add UnknownOption exception, Return None on empty option val

### LINKED ISSUE #7: Throw a useful error if option isn't set
```python
assert s.show_options('default-shell', g=True)
```

```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../libtmux/libtmux/session.py:374: in show_options
    return self.show_option(option, g=g)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Session($1 libtmux_XxTayh), option = [], g = True

    def show_option(self, option, g=False):
        """Return a list of options for the window.

            :todo: test and return True/False for on/off string

            :param option: option to return.
            :type option: string
            :rtype: string, int or bool

            """

        tmux_args = tuple()

        if g:
            tmux_args += ('-g',)

        option = self.cmd(
            'show-options', option, *tmux_args
        ).stdout
>       option = [tuple(item.split(' ')) for item in option][0]
E       IndexError: list index out of range
```

Have a ``raise=False`` argument. If not raising, return ``None``.

tony:
Fixes #7

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,9 @@ Changelog
 
 Here you can find the recent changes to libtmux
 
+- :feature:`8` new exception ``UnknownOption``
+- :feature:`8` return ``None`` for options that are valid tmux options,
+  but unset at that scope.
 - :support:`6` major documentation overhaul
 
 - :release:`0.4.1 <2016-05-23>`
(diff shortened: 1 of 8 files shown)


tony:
See also https://github.com/tony/tmuxp/issues/65

PR #10, opened 2016-07-13
support moving windows to other sessions

dankrause:
Added a `session` argument to `Window.move_window`, and supplied a default empty string argument to the `destination` arg.

This allows you to pass a session ID to `Window.move_window`, which will cause the window to change sessions.

I've also updated the tests to include explicit coverage for `Window.move_window` (with and without session moving) and `Server.new_session`.

### DIFF
--- a/libtmux/window.py
+++ b/libtmux/window.py
@@ -278,19 +278,22 @@ class Window(TmuxMappingObject, TmuxRelationalObject):
 
         self.server._update_windows()
 
-    def move_window(self, destination):
+    def move_window(self, destination="", session=None):
         """Move the current :class:`Window` object ``$ tmux move-window``.
 
         :param destination: the ``target window`` or index to move the window
-            to.
-        :type target_window: string
+            to, default: empty string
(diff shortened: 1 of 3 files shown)


libtmux/window.py
```diff
@@ -278,19 +278,22 @@ def kill_window(self):
...
-    def move_window(self, destination):
+    def move_window(self, destination="", session=None):
         """Move the current :class:`Window` object ``$ tmux move-window``.
 
         :param destination: the ``target window`` or index to move the window
-            to.
-        :type target_window: string
+            to, default: empty string
+        :type destination: string
+        :param session: the ``target sesion`` or index to move the
```
tony: typo

dankrause:
Typo corrected

tony:
Looks good to me.

Do you need this in a project you're working on? I can bump the version on pypi if you want.

dankrause:
No need to bump the pypi version for this small change. It can wait for a release.

PR #11, opened 2016-07-28
allow splitting panes horizontally

tek:
`vertical` parameter for `split_window()`, defaulting to True


tony:
@tek 

Can you add tests for vertical/non-vertical splits?

tek:
there you go!

tony:
merged, do you need this in a project?  I can bump the pypi version up if you wish

tek:
it's going to be a while until I make a release, so don't worry about it

PR #19, opened 2016-12-19
Add start_directory to support to new sessions and panes

gandelman-a:
This allows specifying start_directory for creation of both
session and panes.

Clarifies some docstrings as well.

### DIFF
--- a/libtmux/pane.py
+++ b/libtmux/pane.py
@@ -103,7 +103,7 @@ class Pane(TmuxMappingObject, TmuxRelationalObject):
 
         self.cmd('send-keys', '-R \; clear-history')
 
-    def split_window(self, attach=False, vertical=True):
+    def split_window(self, attach=False, vertical=True, start_directory=None):
         """Split window at pane and return newly created :class:`Pane`.
 
         :param attach: Attach / select pane after creation.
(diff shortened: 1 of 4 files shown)


tony:
Looks good to me, 👍

tony:
@gandelman-a do you need this published? I can tag a new version w/ this PR

gandelman-a:
Hi Tony, no.. thats fine. Just setting up some scripts for my dev. env. and running from master so its fine. Thanks!

tony:
Understood. This got released in 0.6.1 too

tony:
thanks for the contribution again @gandelman-a ! nicely done!

haakenlid:
tmux version 1.8 doesn't support the `tmux new-session -c <start_directory>` option. This caused an error in the tmuxp tests.

https://travis-ci.org/tony/tmuxp/jobs/190761397#L917
`
E           LibTmuxException: [u'tmux: unknown option -- c', u'usage: new-session [-AdDP] [-F format] [-n window-name] [-s session-name] [-t target-session] [-x width] [-y height] [command]']`

A possible workaround is to instead use `default-path` in tmux 1.8 (not supported in later versions) or to use subprocess to run something like this `(cd $start_directory; tmux new-session -t $session_name)`, which should work in all versions.

PR #22, opened 2017-01-14
Fix error when using is_version with lettered tmux version

minijackson:
is_version('something') fails with installed tmux 1.9a

To allow tony/tmuxp#197 ;-)

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -11,6 +11,7 @@ import os
 import re
 import subprocess
 from distutils.version import StrictVersion
+from pkg_resources import parse_version
 
 from . import exc
 from ._compat import console_to_str
@@ -387,7 +388,7 @@ def is_version(version):
 
     installed_version = proc.stdout[0].split('tmux ')[1]
 
-    return installed_version == version
+    return parse_version(installed_version) == parse_version(version)
 
 
 def has_required_tmux_version(version=None):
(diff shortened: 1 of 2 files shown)


tony:
@minijackson Can this have a test for it please?

tony:
Also look at this: https://stackoverflow.com/questions/11887762/compare-version-strings/21065570#21065570

minijackson:
There you go!

But I don't know how to mock the installed tmux to be tmux 1.9a, so while the case that the requested tmux version is "lettered" is tested, the installed tmux case is not.

tony:
Travis automatically tests across those tmux versions.

also you could manually insert the version number for '1.9a'

minijackson:
> also you could manually insert the version number for '1.9a'

I didn't understand, is it something I have to do for this PR ? Where should I insert the version number ?

tony:
That is fine.  

The only thing I'm double checking is how broad support is for `pkg_resources`. Are there any corner cases where the user may not have `setuptools` and this patch could lead to a `ImportError: No module named 'pkg_resources'`.

tony:
Even though most machines are going to have setuptools on it,  after further consideration it may be safer to use a method without it.  I'm not sure what the ramifications would be for upstream packages for tmuxp on arch and slack.

tony:
Try `LooseVersion` instead of `StrictVersion`.

Issue #64, opened 2017-08-08
tmux_cmd doesn't include kwargs


askedrelic:
In tmux_cmd https://github.com/tony/libtmux/blob/189114cde7e125a482ab984749637fa44d66b284/libtmux/common.py#L132

the func receives args and kwargs but only args are added to form the cmd str
https://github.com/tony/libtmux/blob/189114cde7e125a482ab984749637fa44d66b284/libtmux/common.py#L176

Here is some IPDB output showing this, from tracking down https://github.com/tony/tmuxp/issues/184
```
# Reach tmux_cmd() with args and kwargs
> /usr/local/var/pyenv/versions/2.7.11/lib/python2.7/site-packages/libtmux/server.py(99)cmd()
     97                 raise ValueError('Server.colors must equal 88 or 256')
     98
---> 99         return tmux_cmd(*args, **kwargs)
    100
    101     def _list_sessions(self):

ipdb> args
self = <libtmux.server.Server object at 0x10fd11350>
args = [u'-2', 'set-environment', 'FOO', 'bar']
kwargs = {u'-t': u'$8'}

# Reach cmd and kwargs not included
ipdb>
> /usr/local/var/pyenv/versions/2.7.11/lib/python2.7/site-packages/libtmux/common.py(179)__init__()
    177         cmd = [str(c) for c in cmd]
    178
--> 179         self.cmd = cmd
    180
    181         try:

ipdb> cmd
['/usr/local/bin/tmux', '-2', 'set-environment', 'FOO', 'bar']
```

So this seems pretty big; any command using tmux_cmd would apply in the current session instead of the target session. Maybe it hasn't been noticed since a common use case isn't use tmuxp inside of an existing TMUX session?

tony:
``kwargs`` weren't intended to be supported like that in current / prior versions.

That said, I have no objection to adding it.  That said, I did want the signature on ``tmux_cmd`` to be similar to [subprocess](https://docs.python.org/2/library/subprocess.html)' signature. 

To play a bit of devil's advocate in terms of an intuitive, python API: Do you know any other python subprocess-like (command) libraries that accept command line flags as kwargs?

tony:
It does seem big that ``kwargs = {u'-t': u'$8'}`` is being passed into tmux_cmd.

Think we could smoosh that back into `args`?

askedrelic:
[...]

PR #65, opened 2017-08-08
Add session id to Session cmds, when appropriate 

askedrelic:
Starting on fixing #64.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ env:
   - TMUX_VERSION=1.9a
 matrix:
   allow_failures:
+  - python: pypy3.3-5.2-alpha1
   - env: TMUX_VERSION=master
 before_install:
   - export PIP_USE_MIRRORS=true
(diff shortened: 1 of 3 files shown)


tony:
Let's try to take the approach where we smooth out the `*args` (https://github.com/tony/libtmux/issues/64#issuecomment-321177837). We can use `+=` instead of `.extends(` to avoid an extra function call (https://github.com/tony/libtmux/issues/64#issuecomment-321397704).

askedrelic:
So I rebased my code to use `+=` to add the session id, if it is not set. It's a little more complicated due to tmux requiring the `-t` arg be in a specific position, but the test suite should be passing!

askedrelic:
Hmm looks like tests broke due to Py3. I'm still using Py2 as my main environment, I will take a look!

tony:
@askedrelic nice job!

askedrelic:
Hooray, thanks for the merge @tony 🎊

tony:
released in v0.7.4

Issue #67, opened 2017-08-16
Server.has_session() erroneously reports session exists with tmux 2.5


jlargentaye:
tmux from MacPorts (on macOS 10.12.6), recently updated to tmux 2.5.

Tmuxp erroneously reports the session already exists when trying to load the session.

Some light debugging points to libtmux/server.py:Server.has_sessions() does not account for 'tmux has-session' now responds with 'no current session'.

jlargentaye:
After doing a "tmux kill-server", I can't reproduce the issue. It's looks like previously the tmux server had stayed running after the last session finished. 

I'm not immediately finding a way to get it back into that state.

jlargentaye:
Untested fix in pull-request tony/libtmux#68

tony:
https://github.com/tmux/tmux/blob/e802b683ea0ebfa377a5bb8a6b03454c1edbe2d5/cmd-set-environment.c#L76

https://github.com/tmux/tmux/blob/0824850bbce86c395230700cb10c81f162319858/options.c#L639

I can't think of how to recreate it, but that'd be helpful in making a testcase for it.

PR #68, opened 2017-08-17
Make Server.has_session() use returncode

jlargentaye:
has_session() would erroneously return true if tmux returned an unexpected
string, such as "no current session".

Instead of adding yet another string to the list to check against, use the
return code of the 'tmux has-session' command, which is documented to return 0
(Shell true) if targeted session exists, and return 1 (Shell false) in any
other case.

This is **untested**. While it passed selftest, so did the unpatched version.

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -189,6 +189,7 @@ class tmux_cmd(object):
             self.process.stdout.close()
             stderr = self.process.stderr.read()
             self.process.stderr.close()
+            returncode = self.process.returncode
         except Exception as e:
             logger.error(
                 'Exception for %s: \n%s' % (
@@ -196,6 +197,8 @@ class tmux_cmd(object):
                     e
                 )
             )
+
+        self.returncode = returncode
 
         self.stdout = console_to_str(stdout)
         self.stdout = self.stdout.split('\n')
--- a/libtmux/server.py
+++ b/libtmux/server.py
@@ -334,23 +334,10 @@ class Server(TmuxRelationalObject, EnvironmentMixin):
 
         proc = self.cmd('has-session', '-t%s' % target_session)
 
-        if not proc.stdout:
+        if not proc.returncode:
             return True
-        if any(
-            x in proc.stdout for x in
-            ['failed to connect to server', 'error connecting to']
-        ):
-            return False
-        elif 'no server running' in proc.stdout:  # tmux 2.0
-            return False
-        elif 'can\'t find session' in proc.stdout:  # tmux 2.1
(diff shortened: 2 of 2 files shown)


jlargentaye:
Adresses tony/libtmux#67

tony:
That simplification looks nice.

To play devil's advocate / pythonics:

What about keep the string comparisons and raising exceptions? e.g. https://libtmux.git-pull.com/en/latest/api.html#exceptions and https://github.com/tony/libtmux/blob/189114cde7e125a482ab984749637fa44d66b284/libtmux/common.py#L523

Older versions of tmux could still subclass a new command base exception like `TmuxCommandException`.

That way the front-end user could `try:` and `except:` and not have to look up old tmux versions to see what exactly went wrong (since the language changed between tmux versions, one of libtmux' features is abstracting out version differences).

tony:
I need more time on this, because I need to research and assure consistent behavior across older tmux versions.

Does #67 still happen?  Does it happen with all tmux versions, or only certain ones?

In https://github.com/tony/libtmux/issues/67#issuecomment-322926711 it says:

> After doing a "tmux kill-server", I can't reproduce the issue. It's looks like previously the tmux server had stayed running after the last session finished.

Can you try again? Have you ruled out it being a partial/fnmatch of the session name?

tony:
Sorry for the late response, I am having issues keeping up with the project. I'm seeking maintainers @ https://github.com/tony/tmuxp/issues/290. If you (or someone you know) may be interested in helping maintain libtmux or tmuxp, feel free to contact me by email or in the https://github.com/tony/tmuxp/issues/290 issue.

Shados:
@tony the `has-session` command has used the exit code in this manner from right when the command was introduced (I checked), so you don't need to worry about older tmux versions. 

#67 was caused by the relatively rare situation of a tmux session remaining alive with no sessions, likely due to still having an attached client (somehow) and thus causing an unexpected error message when calling `has-session`. You can replicate this trivially using the option recently added to tmux master in tmux/tmux@623f4b12d3af7c6908c6abab1e95589c59f554ef (which is how I ran into the issue, and why I'm commenting).

This PR would fix that issue, and also provide resistance against future changes to `has-session`'s error messages. Which aren't entirely unlikely, given how many different variants you're already accounting for...

tony:
@Shados thank you for the extra info!

@jlargentaye Thanks for the patch!

Looks good so far: https://travis-ci.org/tmux-python/libtmux/builds/352060852

If things go well, we can have this shipped out to 0.8.0, hopefully today.

Issue #77, opened 2017-11-09
Normalize how commands work, debugging


tony:
Restructure how commands are passed in

This will probably be by creating a command factory where the command is built, then an execute() for it. But to the user in a tmux object, ``.cmd`` should still execute immediately, ``_cmd()`` could leave an object where the user can see the exact command tmux is running before hand.

There should be more options to toggle debug settings

PR #79, opened 2017-11-10
Split init / exec of tmux_cmd for debuggability

tony:
_Edit 2022-08-14 by @tony : This is very similar to [`SubprocessCommand`](https://libvcs.git-pull.com/internals/subprocess.html#libvcs._internal.subprocess.SubprocessCommand) ([source](https://github.com/vcs-python/libvcs/blob/v0.14.0/libvcs/_internal/subprocess.py))_

tmux_cmd initialization composes the command

the command will be available in the instance attribute `.cmd`

`.execute()` instance method will run tmux_cmd

See `Server.cmd` for example of new usage

Related #77

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -147,8 +147,7 @@ class EnvironmentMixin:
         return vars_dict
 
 
-class tmux_cmd:
-
+class TmuxCommand:
     """
     :term:`tmux(1)` command via :py:mod:`subprocess`.
 
@@ -188,8 +187,8 @@ class tmux_cmd:
     Notes
     -----
 
-    .. versionchanged:: 0.8
-        Renamed from ``tmux`` to ``tmux_cmd``.
+    .. versionadded:: 0.8.4
+        Wrap to split execution from command from instance of it
     """
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


tony:
This, as of c9598fb *doesn't* touch any of the Server/Session/Window/Pane ``.cmd`` method API's.

As a second part of this, I may add more things to split introspection of commands.

For instance:

- Renaming tmux_cmd to TmuxCommand
- Documenting how they are composed and can be configured, and why/how it's differs from standard library subprocess
- Methods in Tmux objects, e.g. ``attach_session()`` may have a way to return the pure TmuxCommand object (and therefore preview the command generated, including the valuable target information (-t))
- as a way the above could be implemented, it may be via adding methods like ``_attach_session()`` that return TmuxCommand and having ``attach_session`` wrap them and keep normal behavior
- Another thing to consider is to *break* api's by requiring ``.execute()`` be ran to execute commands, and return TmuxCommand by default
  
 - BUT, allow for chaining commands. e.g.,  ``session.attached_window.set_window_option('main-pane-height', 50).select_layout('main-horizontal').execute()``, this would give the equivalent to chaining commands with semicolons. This could substantially speed things up by allowing for composability with commands.

tony:
This is very similar to [`SubprocessCommand`](https://libvcs.git-pull.com/internals/subprocess.html#libvcs._internal.subprocess.SubprocessCommand) ([source](https://github.com/vcs-python/libvcs/blob/v0.14.0/libvcs/_internal/subprocess.py))

tony:
Closing as obsolete. This PR (Sep 2021) split `tmux_cmd.__init__` from execution to enable better debuggability via a new `TmuxCommand` class. The implementation predates the `src/` layout migration and references `libtmux/common.py` and `libtmux/server.py` directly — both files have since been moved to `src/libtmux/`, producing "modify/delete" conflicts on every change.

The underlying goal — separating tmux command construction from execution for debuggability — is still valuable, but the implementation needs to be redone against the current code structure. Master has since added structured logging via `extra={"tmux_cmd": shlex.join(cmd), "tmux_exit_code": ...}` (see `src/libtmux/common.py:264`) which already provides much of the debuggability this PR was after.

Confirmed during a 2026-04-26 batch rebase audit.

PR #120, opened 2018-05-18
Add method `capture_pane` to`Pane`

ritiek:
This new method `Pane.capture_pane` will be very handy for capturing text from a tmux pane.

```python
>>> pane.capture_pane()
['lots and...',
 '..lots of..',
 '.cute text.']
```
So kawaii. :D


tony:
Nice job.

PR #121, opened 2018-05-22
Add optional parameter `literal` to `Pane.send_keys`

### LINKED ISSUE #88: Do not send-keys literally by default
I noticed that libtmux sends keys to a window pane literally. A minimal reproduction:
```python
import libtmux

server = libtmux.Server()
session = server.new_session('happy_session')
window = session.new_window('happy_window')
pane = window.attached_pane

# instead of KeyboardInterrupt, libtmux sends 'C-c'
pane.send_keys('C-c')
```

tmux uses `-l` to specifiy if the keys are to be sent literally:
```bash
# not literally
tmux send-keys -t happy_session C-c

# literally
tmux send-keys -l -t happy_session C-c
```

It would be nice if there were an option to override this behavior in libtmux and preferably set as defaults `Pane.send_keys(literal=False)` just as in tmux.

ritiek:
Closes #88.

This PR adds an optional parameter `literal=False` to `Pane.send_keys` which does not send keys to a tmux pane literally by default. When `literal` is set to `True`, it prefixes the `send-keys` command with `-l` parameter which allows sending keys literally.


tony:
@ritiek Can you rebase this (so the tests again?)

Seanmatthews:
Seems the only error for this was a timeout? Can we rerun the test?

ritiek:
Ok, they passed.

tony:
@ritiek Can we add a test for this?

tony:
@ritiek merged! Thank you!

PR #131, opened 2018-07-09
Fix for unicode commands in Python 2

myw:
Before this PR, if a command was passed into `tmux_command` that had unicode characters in it, such as in the name of a session, format string, or window, a unicode error would be thrown when tmuxp was run with Python 2. The provided unit test shows an example failing test case of this behavior—the real-world use-case is `config.yml` files for tmuxp that contain unicode characters in their names.

This PR resolves the issue by using a python-version-independent conversion inside the `tmux_command` constructor (instead of `str`), assuming that unicode data being sent in Python 2 is encoded as `utf_8`, which is consistent with other places in the code that assumption is made.

Happy to answer any questions, an thanks for making this!

### DIFF
--- a/libtmux/_compat.py
+++ b/libtmux/_compat.py
@@ -92,3 +92,9 @@ else:
 
 
 number_types = integer_types + (float,)
+
+def str_from_console(s):
+    try:
+        return text_type(s)
+    except UnicodeDecodeError:
+        return text_type(s, encoding='utf_8')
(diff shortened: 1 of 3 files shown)


tony:
@myw Sorry for the delay, can you do a rebase?

myw:
@tony Rebased! Thanks!

tony:
@myw Merged! Thanks!

PR #172, opened 2019-01-25
Handle next-X.Y versions

sloria:
tmuxp currently crashes when used with
tmux next-2.9 (the latest Homebrew release).

```
TypeError: '<' not supported between instances of 'str' and 'int'
```

This patch avoids the error by making `get_version()`
strip the "next-" part of the version.

closes tmux-python/tmuxp#449


tony:
@sloria Thank you!

PR #278, opened 2020-05-06
Fix deprecation warning when using collection.MutableMapping

d1618033:
fixes #277

PR #294, opened 2020-07-14
Import MutableMapping from collections.abc instead of collections

Flowdalic:
Importing MutableMapping from collections is deprecated since Python
3.3 and will stop working in 3.9. See the following deprecation notice:

libtmux/common.py:233: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working

Hence we now import from collections.abc.

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -6,7 +6,7 @@ libtmux.common
 ~~~~~~~~~~~~~~
 
 """
-import collections
+import collections.abc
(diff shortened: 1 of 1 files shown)


Flowdalic:
Closing as its a duplicate of https://github.com/tmux-python/libtmux/pull/278

Issue #251, opened 2019-12-25
capture_pane hangs with `-S-`


deliciouslytyped:
I'm trying to expand the `capture_pane` functionality.
I'm using libtmux master on python3.7 with tmux 3.0 -
The following snippet hangs on some panes, while the same command works when run from bash;
```
>>> import libtmux; s = libtmux.Server(); s.children[0].children[0].children[1].cmd("capture-pane", '-p', '-J', '-e', '-E-', '-S-').stdout 
<omitted command noise>
DEBUG: ['/etc/profiles/per-user/myuser/bin/tmux', 'capture-pane', '-t', '%4', '-p', '-J', '-e', '-E-', '-S-']
<no output till ^C>
```
(I added some simple print debugging.)

`list-panes` gives the following:
```
$ tmux list-panes -t 0:
0: [135x17] [history 1952/2000, 939964 bytes] %0
1: [135x17] [history 928/2000, 431099 bytes] %4
2: [135x21] [history 1869/2000, 1194568 bytes] %5
3: [135x21] [history 0/2000, 0 bytes] %7
4: [271x13] [history 1816/2000, 269415 bytes] %1 (active)
5: [271x12] [history 6/2000, 1537 bytes] %132
```

The hang only seems to happen if I pass the `-S-` argument.

P.S.: previously at https://github.com/tmux-python/tmuxp/issues/529

deliciouslytyped:
I took a run at it with the debugger and it seems to hang in https://github.com/tmux-python/libtmux/blob/1331743bcd4ecd6f952ee056570d1462734678b9/libtmux/common.py#L208

deliciouslytyped:
Is this possibly the problem? https://stackoverflow.com/questions/1445627/how-can-i-find-out-why-subprocess-popen-wait-waits-forever-if-stdout-pipe

deliciouslytyped:
The following patch appears to work
```patch
diff --git a/libtmux/common.py b/libtmux/common.py
index 9681d3e..64f4313 100644
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -205,11 +205,7 @@ class tmux_cmd(object):
             self.process = subprocess.Popen(
                 cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE
             )
-            self.process.wait()
-            stdout = self.process.stdout.read()
-            self.process.stdout.close()
-            stderr = self.process.stderr.read()
-            self.process.stderr.close()
[...]

PR #297, opened 2020-09-14
Better interaction with subprocess std[in|out|err]

gil-obradors:
Needed for interact with big buffer.

Thanks to communicate method ~implemented with python > 3.~
~I don't know how to deal with it with python2 ?~

https://docs.python.org/3.5/library/subprocess.html#subprocess.Popen.communicate

#251

Official docs:

    Warning
    Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS   pipe buffers filling up and blocking the child process.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,7 +6,7 @@ Here you can find the recent changes to libtmux
 
 current
 -------
-- :issue:`251`: Enchance subprocess interaction stdin/out/err. Needed for interact with big buffer
+- :issue:`251`: Enchance subprocess interaction std[in|out|err]. Needed for interact with big buffer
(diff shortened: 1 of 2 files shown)


gil-obradors:
Official docs says:

    Warning
    Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS   pipe buffers filling up and blocking the child process.

gil-obradors:
Is ok in python2 and 3!

tony:
@gil-obradors Merged!   This will be in 0.8.5

PR #298, opened 2020-09-24
split on a special wildcard instead of tab

JonathanRaiman:
Several APIs in libtmux fail with Tmux 3.1 installed but worked with Tmux 3.0/3.0a.
It appears that the format string used by the library to communicate with Tmux will ignore tabs and return lowercases instead. I switch the tab to a wildcard (e.g. `$@$`) but this could be any kind of string assuming it's sufficiently rare to not collide with desired user input.

PR #305, opened 2020-12-06
fix: use `|` instead of `\t` for format sepatator

jagguli:
- for some reason on my arch install tmux does not print back tabs

I'm not sure if it affects other but just puttin this out there

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.x"]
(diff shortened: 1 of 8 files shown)


tony:
@jagguli This interests me.

I'm interested in perhaps have a `SEP` constant changeable by an env variable.

I'm not sure if this will effect other things by side effects.  Other people have different systems, how can we be sure `|` doesn't break things for others?

tony:
Any difference between this and #298 ?

Can you give this a rebase? 

cc:  @JonathanRaiman

jagguli:
rebased in https://github.com/tmux-python/libtmux/pull/343

PR #341, opened 2021-10-23
Fix crash on calling Server.attached_sessions

Timoses:
fixes #180

Issue #180, opened 2019-02-12
Possible bug in attached_sessions() method


amitibo:
Calling the `attached_sesssions` method of a `Server` object raises the following exception:
```
In [20]: import libtmux

In [21]: server = libtmux.Server()

In [22]: server.attached_sessions
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-22-aee9c49879f2> in <module>()
----> 1 server.attached_sessions

/dccstor/faceid/code/anaconda3/x86_64/envs/mxnet/lib/python2.7/site-packages/libtmux/server.pyc in attached_sessions(self)
    336             if 'session_attached' in session:
    337                 # for now session_active is a unicode
--> 338                 if session.attached == '1':
    339                     logger.debug('session %s attached', session.name)
    340                     attached_sessions.append(session)

AttributeError: 'dict' object has no attribute 'attached'
```
Seems that `line 338` above expects an object of type `Session` while it is actually a `dict`.
Using `libtmux=0.81` and `python=2.7`

Jim-Shaddix:
I encountered the same issue. I am thinking that code block should look something like this. I am not really sure though. At the very least it stops the errors from occuring, and will return a Session object. 
```
            if 'session_attached' in session:
                # for now session_active is a unicode
                if session['session_attached'] == '1':
                    logger.debug('session %s attached', session['session_name'])
                    attached_sessions.append(session)
```

rajaravivarma-r:
If you don't want to modify the library code, you can do this instead `attached_session = next(s for s in server.list_sessions() if s['session_attached'] == '1')`

PR #342, opened 2021-10-23
Remove attached sessions limitation to not detect multiple attached clients

Timoses:
Technically this is a breaking change. Previously one could expect to detect that only **one** client is attached to the sessions via this property.

However, since previously `Server.attached_sessions` was broken (see #180 and PR to fix #341) I do not expect anybody to use the property currently.

Builds on https://github.com/tmux-python/libtmux/pull/341/files (<- Merge first)

### DIFF
--- a/libtmux/server.py
+++ b/libtmux/server.py
@@ -330,13 +330,13 @@ class Server(TmuxRelationalObject, EnvironmentMixin):
         attached_sessions = list()
 
         for session in sessions:
-            if 'session_attached' in session:
-                # for now session_active is a unicode
-                if session.attached == '1':
-                    logger.debug('session %s attached', session.name)
-                    attached_sessions.append(session)
-                else:
-                    continue
+            attached = session.get('session_attached')
+            # for now session_active is a unicode
+            if attached == '1':
+                logger.debug('session %s attached', session.get('name'))
+                attached_sessions.append(session)
+            else:
+                continue
 
         return [Session(server=self, **s) for s in attached_sessions] or None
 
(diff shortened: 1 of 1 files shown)


tony:
@Timoses Can you rebase this and #341 so tests run? Thank you!

Timoses:
@tony : Done ; )

tony:
@Timoses Thank you!

libtmux/server.py
```diff
@@ -319,8 +319,6 @@ def attached_sessions(self):
         """
         Return active :class:`Session` objects.
 
-        This will not work where multiple tmux sessions are attached.
```
tony, line 322: Hm, back in history, I'm wondering what I will thinking back then.

https://github.com/tmux-python/tmuxp/commit/1519deeed2d5299e34eb48d4b530e48bcde939db

I guess there really wasn't a good rationale back then.

tony:
@Timoses Thank you!

Would you like me to cut a new release of libtmux for this now?

tmuxp?

tony:
Also does this look good? I merged #342 directly but preserved #341's commit.

tony:
@Timoses This is live in `0.10.2` if you'd like to try!

Timoses:
> Also does this look good? I merged #342 directly but preserved #341's commit.

Looks good.

> @Timoses This is live in 0.10.2 if you'd like to try!

Thanks, already tried it and it works : ).

What I noticed though is that using more tmux introspection causes more "lag". As I am using libtmux in a real-time application I am now circumventing calling it too often: https://codeberg.org/timoses/caster-timoses/src/branch/master/caster_timoses/tmux/__init__.py#L19-L26

PR #348, opened 2022-01-11
Trying to get a compliant, typed version parser/comparator

tony:
Companion: https://github.com/tmux-python/tmuxp/pull/727

See also:
- https://github.com/asottile/flake8-typing-imports/blob/923a533/flake8_typing_imports.py#L32-L42
- https://github.com/pypa/packaging/blob/5984e3b25/packaging/version.py#L444

- [x] String comparison's against `Version`

   @layday's [snippet](https://github.com/pypa/packaging/issues/465#issuecomment-1013715662)
- [ ] LegacyVersion deprecation 

  In [`packaging`](https://packaging.pypa.io/), `LegacyVersion` (the supercessor to `LooseVersion`) will be deprecated, we won't be able to support `2.4-openbsd` (like what is seen on OpenBSD's tmux versions

  ```
    /home/t/work/python/libtmux/.venv/lib/python3.10/site-packages/packaging/version.py:127: 
      DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
      warnings.warn(
  
  -- Docs: https://docs.pytest.org/en/stable/warnings.html
  ```
- [ ] tmux versions

       To the above, consider subclassing `Version` for tmux versions to handle git builds, platform builds, etc. then keeping `Version` for supporting pypi releases
- [ ] `Version` Assure that `libtmux.__version__` / `tmuxp.__version__` supports git refs / `local` match groups

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -6,12 +6,13 @@ libtmux.common
 
 """
 import collections
+import functools
 import logging
 import os
 import re
 import subprocess
 import sys
-from distutils.version import LooseVersion
+import typing
 
 from . import exc
 from ._compat import MutableMapping, console_to_str, str_from_console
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Hi @Tony
So in response to your question in the yt issue, I'm actually not sure my solution `version_tuple` fits your needs, if you want to support version schemes that used to rely on `distutils.LooseVersion`, my function may not work. 

`version_tuple` is explicitly aimed only at parsing int tuples, and will truncate any trailing str to support alphas/betas

for instance
```python
>>> version_tuple("2.13-hello")
(2, 13)
```
is this really what you need ?

tony:
@neutrinoceros I saw your follow up [here](https://github.com/yt-project/yt/pull/3604#issuecomment-1013704332) and realize it is explicitly tuple related and I'm dabbling with keeping it as an object: #351. I've underestimated the effort required.

I suppose if it takes no time at all, you could PR the function in src/libtmux/util.py

neutrinoceros:
Seems like you got a better solution from the packaging thread, I won't open a PR after all. Anyway feel free to copy my solution in case you need it still.

tony:
> Seems like you got a better solution from the packaging thread, I won't open a PR after all. Anyway feel free to copy my solution in case you need it still.

Thank you @neutrinoceros!  @layday's [mixin](https://github.com/pypa/packaging/issues/465#issuecomment-1013715662) for `Version` seems quite nice and I'm trialing it in another PR here.

In case anyone googles their way here:

```python
from functools import total_ordering

from packaging.version import Version


@total_ordering
class _VersionCmpMixin:
    def __eq__(self, other: object) -> bool:
        if isinstance(other, str):
            other = self.__class__(other)
        return super().__eq__(other)

    def __lt__(self, other: object) -> bool:
        if isinstance(other, str):
            other = self.__class__(other)
        return super().__lt__(other)


class MyVersion(_VersionCmpMixin, Version):
    pass


assert MyVersion("3") == "3.0.0"
assert MyVersion("3") != "3.0.1"
```

PR #358, opened 2022-03-10
docs: Code prompt (fix this later)

tony:
later in evening / weekend


tony:
Unneeded  https://github.com/tmux-python/libtmux/blob/2db8bba4ed37efe653a32ced0b6e68f053cf581a/docs/conf.py#L100-L102

PR #359, opened 2022-03-18
Adding unobtrusive type hinting

otherJL0:
Adding simple type hints in independent functions as a first step to introduce type hints to the project

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -10,8 +10,8 @@ import os
 import re
 import subprocess
 import sys
+import typing as t
 from distutils.version import LooseVersion
-from typing import Optional
 
 from . import exc
 from ._compat import MutableMapping, console_to_str, str_from_console
@@ -376,7 +376,7 @@ class TmuxRelationalObject:
 
 def which(
     exe: str,
-    default_paths: list[str] = [
+    default_paths: t.List[str] = [
         "/bin",
         "/sbin",
         "/usr/bin",
@@ -384,7 +384,7 @@ def which(
         "/usr/local/bin",
     ],
     append_env_path: bool = True,
-) -> Optional[str]:
+) -> t.Optional[str]:
     """
     Return path of bin. Python clone of /usr/bin/which.
 
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -11,6 +11,7 @@ import re
 import subprocess
 import sys
 from distutils.version import LooseVersion
+from typing import Optional
 
 from . import exc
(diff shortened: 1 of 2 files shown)


libtmux/common.py
```diff
@@ -374,10 +375,16 @@ def get_by_id(self, id):
 
 
 def which(
-    exe=None,
-    default_paths=["/bin", "/sbin", "/usr/bin", "/usr/sbin", "/usr/local/bin"],
-    append_env_path=True,
-):
+    exe: str,
+    default_paths: list[str] = [
```
tony, line 379: Needs 3.9: https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

we can either wait until Monday (when i drop python 3.7 and 3.8 and you can rebase) or use `typing.List`

tony: if you change is to `typing.List` and CI passes I can merge this immediately

otherJL0: Sure, no problem! Is there a style preference between the options below?

```python
# Import all type hints at top that will be used in this file, may be very long
from typing import Optional, List, Union, Any

def sample(array: List[Any], id: Union[str, int]) -> Optional[int]:
    ...
    return None
```

```python
# Use an alias to avoid explicitly importing types
import typing as T

def sample(array: T.List[T.Any], id: T.Union[str, int]) -> T.Optional[int]:
    ...
    return None
```

otherJL0: I realize that I took the first approach by importing `Union` and `Optional` but that should be a quick fix if the second style is preferred.

tony:
```
# Use an alias to avoid explicitly importing types
import typing as T
```

Make the `T` lowercase and that would be good

https://cs.github.com/?scopeName=All+repos&scope=&q=%22import+typing+as+T%22

Seems flask does it. https://github.com/pallets/flask/blob/cb7cd1e79b280d3fd80978ca82ae52881525a476/src/flask/blueprints.py

tony (approved):
LGTM

This will make it into 0.11, which will release tomorrow

afterward 3.7 and 3.8 will be dropped

@otherJL0 this Saturday i will be adding type annotations and data structures #360.  If you're around and have any ideas/feedback/PR sugggestions you'd like to contribute feel free

tony:
@otherJL0 libtmux [v0.13.0](https://libtmux.git-pull.com/history.html#libtmux-0-13-0-2022-08-05) has typings, if you'd like to try.

PR #361, opened 2022-03-18
Adding type hints to class methods returning self and child

otherJL0:
* Adding `SessionDict`, `WindowDict`, and `PaneDict` type aliases to `common.py` for _list methods
* Adding hints for methods returning `self` and collections of children

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -25,6 +25,10 @@ TMUX_MIN_VERSION = "1.8"
 #: Most recent version of tmux supported
 TMUX_MAX_VERSION = "2.4"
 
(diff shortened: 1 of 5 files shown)


otherJL0:
@tony I appreciate the invitation! I don't know if I'll have availability to collaborate simultaneous on Saturday but I'm happy to help gradually add type hints when I can.

tony:
@otherJL0 Absolutely, open up a PR / issue / discussion whenever you have free time. We are below < 1.0 so there is flexibility to make changes. 

You will this commit in 0.11

tony (approved):
LGTM

PR #372, opened 2022-05-17
Modify retry() to take a callable and enforce the timeout

### LINKED ISSUE #368: `retry(seconds=...)` broke
Instead

```
import time
t = time.process_time()

while (time.process_time() - t) * 1000 < 2:
    # stuff happens
    time.sleep(0.1)
```

Find cross platform way to do timeout with signals

See also https://github.com/tmux-python/tmuxp/issues/620

categulario:
Previous implementation is equivalent to `return True`. This implementation, while API incompatible, is easy to use (I would modify the corresponding code in tmuxp myself) and enforces the timeout via an exception, which would help debugging since it provide the specific cause.

API docs updated accordingly. Environment variable preserved.

fixes #368 and addresses https://github.com/tmux-python/tmuxp/issues/704#issuecomment-1128288348 , it would also help with https://github.com/tmux-python/tmuxp/issues/620

### DIFF
--- a/libtmux/test.py
+++ b/libtmux/test.py
@@ -5,7 +5,7 @@ import os
 import tempfile
 import time
 import warnings
-from typing import Optional
+from typing import Callable, Optional
 
 from .exc import WaitTimeout
 
@@ -57,12 +57,12 @@ def retry(seconds: Optional[float] = RETRY_TIMEOUT_SECONDS) -> bool:
 
 
 def retry_until(
-    fun,
-    seconds=RETRY_TIMEOUT_SECONDS,
+    fun: Callable,
+    seconds: float = RETRY_TIMEOUT_SECONDS,
     *,
-    interval=RETRY_INTERVAL_SECONDS,
-    raises=True,
-):
+    interval: Optional[float] = RETRY_INTERVAL_SECONDS,
+    raises: Optional[bool] = True,
+) -> bool:
     """
     Retry a function until a condition meets or the specified time passes.
 
@@ -71,7 +71,7 @@ def retry_until(
     fun : callable
         A function that will be called repeatedly until it returns ``True``  or
         the specified time passes.
-    seconds : int
+    seconds : float
         Seconds to retry. Defaults to ``8``, which is configurable via
         ``RETRY_TIMEOUT_SECONDS`` environment variables.
     interval : float
--- a/libtmux/test.py
+++ b/libtmux/test.py
@@ -79,6 +79,10 @@ def retry_until(
     ...     return p.current_path == pane_path
     ...
     ... retry(f)
+
+    In pytest:
+
+    >>> assert retry(f, raises=False)
     """
     ini = time.time()
 
@@ -88,8 +92,9 @@ def retry_until(
             if raises:
                 raise WaitTimeout()
             else:
-                break
+                return False
         time.sleep(interval)
+    return True
 
 
 def get_test_session_name(server, prefix=TEST_SESSION_PREFIX):
--- a/libtmux/test.py
+++ b/libtmux/test.py
@@ -5,10 +5,13 @@ import os
 import tempfile
 import time
 
+from .exc import WaitTimeout
+
 logger = logging.getLogger(__name__)
 
 TEST_SESSION_PREFIX = "libtmux_"
 RETRY_TIMEOUT_SECONDS = int(os.getenv("RETRY_TIMEOUT_SECONDS", 8))
+RETRY_INTERVAL_SECONDS = float(os.getenv("RETRY_INTERVAL_SECONDS", 0.05))
 
 namer = tempfile._RandomNameSequence()
 current_dir = os.path.abspath(os.path.dirname(__file__))
@@ -41,6 +44,52 @@ def retry(seconds=RETRY_TIMEOUT_SECONDS):
     ...         break
     """
     return (lambda: time.time() < time.time() + seconds)()
+
+
+def retry_until(
+    fun,
+    seconds=RETRY_TIMEOUT_SECONDS,
+    *,
+    interval=RETRY_INTERVAL_SECONDS,
+    raises=True,
+):
+    """
+    Retry a function until a condition meets or the specified time passes.
+
+    Parameters
+    ----------
+    fun : callable
+        A function that will be called repeatedly until it returns ``True``  or
+        the specified time passes.
+    seconds : int
+        Seconds to retry. Defaults to ``8``, which is configurable via
+        ``RETRY_TIMEOUT_SECONDS`` environment variables.
+    interval : float
+        Time in seconds to wait between calls. Defaults to ``0.05`` and is
+        configurable via ``RETRY_INTERVAL_SECONDS`` environment variable.
+    raises : bool
+        Wether or not to raise an exception on timeout. Defaults to ``True``.
+
+    Examples
+    --------
+
+    >>> def f():
+    ...     p = w.attached_pane
+    ...     p.server._update_panes()
+    ...     return p.current_path == pane_path
+    ...
+    ... retry(f)
+    """
+    ini = time.time()
+
+    while not fun():
+        end = time.time()
+        if end - ini >= seconds:
+            if raises:
+                raise WaitTimeout()
+            else:
+                break
+        time.sleep(interval)
 
 
 def get_test_session_name(server, prefix=TEST_SESSION_PREFIX):
--- a/libtmux/test.py
+++ b/libtmux/test.py
@@ -4,6 +4,7 @@ import logging
 import os
 import tempfile
 import time
+import warnings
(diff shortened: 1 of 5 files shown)


tony:
@categulario `poetry run flake8` and fix those, and it should let the rest of the tests run

libtmux/test.py
```diff
@@ -34,13 +34,19 @@ def retry(seconds=RETRY_TIMEOUT_SECONDS):
...
+    ...
+    ... retry(f)
     """
-    return (lambda: time.time() < time.time() + seconds)()
+    ini = time.time()
+
+    while not fun():
+        end = time.time()
+        if end - ini >= seconds:
+            raise Exception('Function timed out without meeting condition')
```
tony, line 49: I feel a bit resistant of moving it to a callback. It may be the only choice.

We have many `while retry()` in `test_workspacebuilder.py` ([link](https://github.com/tmux-python/tmuxp/blob/b0e744ec4d8ae83bf877f2c62a3334773970828f/tests/test_workspacebuilder.py)).  This turns the `while` block of code into a callback.

I'd like to see what other options are out there so we're not using callbacks, if it's reasonable to avoid it.

Any chance you could also find a `with` compatible one? e.g. [interruptingcow](https://bitbucket.org/evzijst/interruptingcow/src/master/) ([src](https://bitbucket.org/evzijst/interruptingcow/src/master/interruptingcow/__init__.py))

I'm happy to keep this function as well (in case we want to use this style of invocation or ultimately move to it)

categulario:
In the latest commits I even added three tests

libtmux/test.py
```diff
@@ -5,26 +5,43 @@
...
+RETRY_INTERVAL_SECONDS = float(os.getenv("RETRY_INTERVAL_SECONDS", 0.05))
 
 namer = tempfile._RandomNameSequence()
 current_dir = os.path.abspath(os.path.dirname(__file__))
 example_dir = os.path.abspath(os.path.join(current_dir, "..", "examples"))
 fixtures_dir = os.path.realpath(os.path.join(current_dir, "fixtures"))
 
 
-def retry(seconds=RETRY_TIMEOUT_SECONDS):
+def retry(
```
tony, line 22: Can we:

- keep the old function (so we can deprecate it over the next release)
- rename this new function something like: `until`, `wait`, `retry`, (any other name you find suitable will do)

thoughts?

categulario: sure thing

libtmux/test.py
```diff
@@ -43,6 +46,52 @@ def retry(seconds=RETRY_TIMEOUT_SECONDS):
...
+):
+    """
+    Retry a function until a condition meets or the specified time passes.
+
+    Parameters
+    ----------
+    fun : callable
+        A function that will be called repeatedly until it returns ``True``  or
+        the specified time passes.
+    seconds : int
```
tony, line 64:
```suggestion
    seconds : float
```

Based on [`time.sleep()`](https://github.com/python/typeshed/blob/740193a/stdlib/time.pyi#L76) annotation

libtmux/test.py
```diff
@@ -43,6 +46,52 @@ def retry(seconds=RETRY_TIMEOUT_SECONDS):
     return (lambda: time.time() < time.time() + seconds)()
 
 
+def retry_until(
+        fun,
+        seconds=RETRY_TIMEOUT_SECONDS,
```
tony, line 51:
```suggestion
        seconds: float=RETRY_TIMEOUT_SECONDS,
```

Based on [`time.sleep()`](https://github.com/python/typeshed/blob/740193a/stdlib/time.pyi#L76) annotation

libtmux/test.py
```diff
@@ -43,6 +46,52 @@ def retry(seconds=RETRY_TIMEOUT_SECONDS):
     return (lambda: time.time() < time.time() + seconds)()
 
 
+def retry_until(
+        fun,
```
tony, line 50: Can add [`typing.Callable`](https://docs.python.org/3/library/typing.html#typing.Callable)

libtmux/test.py
```diff
@@ -43,6 +46,52 @@ def retry(seconds=RETRY_TIMEOUT_SECONDS):
...
+        seconds=RETRY_TIMEOUT_SECONDS,
+        *,
+        interval=RETRY_INTERVAL_SECONDS,
```
tony, line 53: Can add `float` per [here](https://github.com/tmux-python/libtmux/pull/372/files#r877933606)

tony:
@categulario 

- Original: b8c2569
- Rebased via github: c041514
- My squash / interactive rebase via `rebase -i`: 036642e

libtmux/test.py
tony: > I gave it some thought in terms of a context manager, but honestly I don't think it is a good idea to add that much complexity (and its [own bugs](https://bitbucket.org/evzijst/interruptingcow/issues?status=new&status=open)) to such a simple use case. I do think spawning a thread/task/context/whatever and reentrantly time it out is actually a bad idea that will complicate debugging.

I think the solution we've come to makes the most sense. I've seen solutions like interruptingcow's on stackoverflow and found the lack of compatibility also a bit troubling.

Would be awesome of python standard library had cross-platform, sync + async, bullet proof solution for this. But it may not be that easy to solve :)

> I have to say that my proposed code is intended to replace the current testing-only retry() function that is broken, but not intended for using in user code like plugins, or tmuxp code itself.

This also is fine with me.

tony:
v0.12.0a0: [branch](https://github.com/tmux-python/libtmux/tree/v0.12.0a0), [pypi](https://pypi.org/project/libtmux/0.12.0a0/)

tony:
@categulario merged in

v0.12.0a0: [branch](https://github.com/tmux-python/libtmux/tree/v0.12.0a0), [pypi](https://pypi.org/project/libtmux/0.12.0a0/)

I will need to drop python 3.7 and 3.8 in tmuxp since that's happening in libtmux before 0.12.0a0 can be added to tmuxp: https://github.com/tmux-python/tmuxp/pull/763

tony:
i will update this

- [x] tmuxp v1.12.0a0: Drop 3.7 + 3.8 for libtmux compatibility:  [branch](https://github.com/tmux-python/tmuxp/tree/v1.12.0a0) [pypi](https://pypi.org/project/tmuxp/1.12.0a0/)
- [ ] tmuxp v1.12.0a1: Update libtmux package for `retry_until`

tony:
https://github.com/tmux-python/tmuxp/pull/776 has a concept of `retry_until()` with `assert retry_until(..., raises=False)`

tony:
@categulario As a next step you are welcome to make a PR moving `retry()` -> `retry_until()` in tmuxp.

if you do have something started let me know, if not i will get to it during the weekend.  I have an example at https://github.com/tmux-python/tmuxp/pull/776

categulario:
> I will need to drop python 3.7 and 3.8 in tmuxp since that's happening in libtmux before 0.12.0a0 can be added to tmuxp: https://github.com/tmux-python/tmuxp/pull/763

Is that necessary? python 3.7 is still maintained. and 3.8 comes in ubuntu 20.04. What exactly makes libtmux incompatible?

> @categulario As a next step you are welcome to make a PR moving retry() -> retry_until() in tmuxp.

Sure thing!

tony:
> Is that necessary? python 3.7 is still maintained. and 3.8 comes in ubuntu 20.04. What exactly makes libtmux incompatible?

I thought about that.

I can look at reversing it, but I'd need to do it in libtmux first.  See #374

tony:
@categulario libtmux and tmuxp support python 3.7 and 3.8 again ✅ 

- [x] libtmux v0.12.0a1 [branch](https://github.com/tmux-python/libtmux/tree/v0.12.0a1), [pypi](https://pypi.org/project/libtmux/0.12.0a1/)
- [x] tmuxp v1.12.0a2 [branch](https://github.com/tmux-python/tmuxp/tree/v1.12.0a2), [pypi](https://pypi.org/project/tmuxp/1.12.0a2/)

Do you still want to move the tests over to `retry_until()` (if so do you have an ETA?) If not I can try to do it in the weekend

P.S. It's open source, none of us our paid, but the reason I press somewhat is the warnings 😄 :
image

categulario:
What is causing the warnings? I'm still interested in doing it. My eta
would be the night of the Saturday which for you is the morning of the
Saturday. I'm in UTC+8

tony:
> What is causing the warnings?
The warnings are from `retry()` being used in `test_workspacebuilder.py` (which should be resolved by moving to `retry_until()`

P.S. And the remainder - in python 3.10 - are from the user `distutils.Version` (which I'll get around to fixing soon)

> I'm still interested in doing it. My eta would be the night of the Saturday which for you is the morning of the Saturday. I'm in UTC+8

That sounds good to me

Issue #396, opened 2022-08-11
libtmux is picking up the wrong tmux version


mdeguzis:
See: https://github.com/tmux-python/tmuxp/issues/794

Started new server, protocol mismatch. tmuxp seems to be using /usr/bin/tmux instead of what I have aliased.

```
$ tmux -V
tmux.orig 2.8

$ /usr/bin/tmux -V
tmux 1.8
```

tmuxp:
```
-------------------------
environment:
        dist: Linux-5.4.204-124.362.amzn2int.x86_64-x86_64-with-glibc2.2.5
        arch: x86_64
        uname: Linux; dev-dsk-deguzim-1a-fa82568b.us-east-1.amazon.com; 5.4.204-124.362.amzn2int.x86_64
        version: #1 SMP Wed Jul 13 03:24:31 UTC 2022
-------------------------

tmux version: 1.8
libtmux version: 0.12.0
tmuxp version: 1.12.1
tmux path: /usr/bin/tmux
tmuxp path: /home/deguzim/.local/lib/python3.7/site-packages/tmuxp
shell: /bin/zsh
-------------------------
tmux sessions:

        protocol version mismatch (client 7, server 8)
tmux windows:

        protocol version mismatch (client 7, server 8)
tmux panes:

        protocol version mismatch (client 7, server 8)
tmux global options:

        protocol version mismatch (client 7, server 8)
tmux window options:

        protocol version mismatch (client 7, server 8)
```

mdeguzis:
Simple test. What I think needs to happen, is the base folder of `which tmux` needs to be appended to the front of the paths:
```
241     def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
242         tmux_bin = which(
243             "tmux",
244             default_paths=kwargs.get(
245                 "tmux_search_paths",
246                 ["/apollo", "/bin", "/sbin", "/usr/bin", "/usr/sbin", "/usr/local/bin"],
247             ),
248             append_env_path=kwargs.get("append_env_path", True),
249         )
```

Will you accept a CR for this?

mdeguzis:
Fixed tmuxp:
```
$ tmuxp freeze --force -f yaml --yes --quiet
Save to: /home/deguzim/.tmuxp/work.yaml [/home/deguzim/.tmuxp/work.yaml]: 
```

Submitted PR: https://github.com/tmux-python/libtmux/pull/397

tony:
[...]

PR #397, opened 2022-08-11
Strict detection of tmux version via regex

mdeguzis:
See: #396 

This change uses a more precise regex method to grab the tmux version string, taking into account the many variations of tmux versions:

```
# tmux 3.3a · tmux 3.3 · tmux 3.2a · tmux 3.2 · tmux 3.1c · tmux 3.1b · tmux tmux.org 2.8
```

I tested my change locally with tmuxp as well:

```
[dev-dsk-deguzim-1a-fa82568b DeGuzim-Config]$ tmuxp freeze --force -f yaml --yes --quiet
Save to: /home/deguzim/.tmuxp/work.yaml [/home/deguzim/.tmuxp/work.yaml]: 
[dev-dsk-deguzim-1a-fa82568b DeGuzim-Config]$ cat ~/.tmuxp/work.yaml 
session_name: work
windows:
- focus: 'true'
  layout: 61a9,231x42,0,0[231x16,0,0{150x16,0,0,0,80x16,151,0,2},231x25,0,17,1]
  options: {}
  panes:
  - shell_command:
    - cd /local/home/deguzim/src/libtmux
    - zsh
  - shell_command:
    - cd /local/home/deguzim/src/tmuxp
    - zsh
  - focus: 'true'
    shell_command:
    - cd /local/home/deguzim/src/DeGuzim-Config
    - python3
  window_name: zsh
```

Testing with python3:
```
>>> ver = "tmux 1.8"
>>> import re
>>> re.search('[0-9]{1}.[0-9]', ver).group(0)
'2.8'

>>> ver = "tmux 2.3"
>>> re.search('[0-9]{1}.[0-9]', ver).group(0)
'2.3'
```
Regex 101 test: https://regex101.com/r/y7jsSJ/1

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -552,7 +552,7 @@ def get_version() -> LooseVersion:
             )
         raise exc.VersionTooLow(proc.stderr)
 
-    version = proc.stdout[0].split("tmux ")[1]
+    version = re.search('[0-9]{1}.[0-9]', proc.stdout[0]).group(0)
 
     # Allow latest tmux HEAD
     if version == "master":


mdeguzis:
hold up, have to patch server.py as well...

mdeguzis:
I think the server that is starting is my /usr/bin/tmux, which is invalid vs what I have aliased. Checking that out...

libtmux/common.py
```diff
@@ -552,7 +552,7 @@ def get_version() -> LooseVersion:
             )
         raise exc.VersionTooLow(proc.stderr)
 
-    version = proc.stdout[0].split("tmux ")[1]
+    version = re.search('[0-9]{1}.[0-9]', proc.stdout[0]).group(0)
```
tony, line 555: What about line 561, wouldn't this lose access to the **a** in **3.2a**?

tony, line 555: `poetry run mypy .`

`libtmux/common.py:555: error: Item "None" of "Optional[Match[str]]" has no attribute "group"`

This will need null-checks to handle empty matches.  I haven't tested, but it'd be something like

```suggestion
    match = re.search('[0-9]{1}.[0-9]', proc.stdout[0])
    assert match is not None
    version = match.group(0)
```

tony (changes requested):
Looking good so far. Good catch and nice job.

1. We may want to for `get_version()` test 3.2a and 3.2
2. I have a regex here: https://github.com/tmux-python/libtmux/blob/d4c9af5/tests/test_common.py#L32, interestingly I don't fetch the letter either

tony:
@mdeguzis I will give your PR priority review if you have more commits

I can also take a stab at it (but would be happy if yours got in so you get credit!)

mdeguzis:
I should be able to look at the comments this week and see what I can do.

tony:
@mdeguzis Is this a PR you're still interested in?

PR #405, opened 2022-08-22
fix(tmux_cmd): remove only trailing new lines in stdout

### LINKED ISSUE #402: tmux_cmd remove empty lines when it should not
Hi,

I'm in the middle of writing a pytest-tmux plugin.
While writing unit/functional testing, I had an assertion test who failed for an obscure reason and didn't get why at first.
I finally discovered that capture-pane doesn't return what it really displayed.............

There's IMO a huge bug [here](https://github.com/tmux-python/libtmux/blob/master/libtmux/common.py#L273).

Yes the empty lines should be cleaned but not all.

For example, a capture-pane of :
```
$ printf "  Hello World  .\n\n"
  Hello World  .

$






```

Should return :
```
$ printf "  Hello World  .\n\n"
  Hello World  .

$
```

and not
```
$ printf "  Hello World  .\n\n"
  Hello World  .
$
```

Any objection to clean empty lines starting from the last and stop until something is found ?

Regards,

rockandska:
Close #402

Seems the branch v0.8x did not exist anymore to backport this change.

### DIFF
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -270,16 +270,21 @@ class tmux_cmd:
 
         stdout_str = console_to_str(stdout)
         stdout_split = stdout_str.split("\n")
-        stdout_filtered = list(filter(None, stdout_split))  # filter empty values
+        # remove trailing newlines from stdout
+        for v in reversed(stdout_split):
+            if v == '':
+                stdout_split.pop()
+            else:
+                break
 
         stderr_str = console_to_str(stderr)
         stderr_split = stderr_str.split("\n")
         self.stderr = list(filter(None, stderr_split))  # filter empty values
 
-        if "has-session" in cmd and len(self.stderr) and not stdout_filtered:
+        if "has-session" in cmd and len(self.stderr) and not stdout_split:
             self.stdout = [self.stderr[0]]
         else:
-            self.stdout = stdout_filtered
+            self.stdout = stdout_split
 
         logger.debug("self.stdout for {}: \n{}".format(" ".join(cmd), self.stdout))
 
--- a/libtmux/common.py
+++ b/libtmux/common.py
@@ -271,11 +271,8 @@ class tmux_cmd:
         stdout_str = console_to_str(stdout)
         stdout_split = stdout_str.split("\n")
         # remove trailing newlines from stdout
-        for v in reversed(stdout_split):
-            if v == '':
-                stdout_split.pop()
-            else:
-                break
+        while stdout_split and stdout_split[-1] == "":
+            stdout_split.pop()
 
         stderr_str = console_to_str(stderr)
         stderr_split = stderr_str.split("\n")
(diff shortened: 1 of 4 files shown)


libtmux/session.py
```diff
@@ -198,10 +198,10 @@ def rename_session(self, new_name: str) -> "Session":
     def new_window(
         self,
         window_name: t.Optional[str] = None,
-        start_directory: None = None,
+        start_directory: str = "",
         attach: bool = True,
         window_index: str = "",
-        window_shell: None = None,
+        window_shell: str = "",
```
tony, line 204: Can you explain these changes?

rockandska: Without fixing this, mypy complained about

```
tests/test_pane.py:72: error: Argument "window_shell" to "new_window" of "Session" has incompatible type "str"; expected "None"
```

If I had specified start_directory, the problem would have been the same, so fixed it at the same time

rockandska: @tony : what do you want me to do ?

Only fix window_shell typing with `t.Optional[str] = None` and fix the others later ?

tony: >  After re-reading the method and its docstring maybe I should have used t.Optional[str] as for window_name and update the docstring too. Let me know

Yes! This would be good

> What do you want me to do ?

Sorry for the delay

> Only fix `window_shell typing with `t.Optional[str] = None and fix the others later ?

I would prefer `t.Optional[str]`, the reason why is it'd preserve runtime behavior.

rockandska: > Sorry for the delay

NP ^^

Squashed commits
Rebased with master
Fixed correct type only for window_shell and add `optional` ind docstring

libtmux/common.py
```diff
@@ -270,16 +270,21 @@ def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
 
         stdout_str = console_to_str(stdout)
         stdout_split = stdout_str.split("\n")
-        stdout_filtered = list(filter(None, stdout_split))  # filter empty values
+        # remove trailing newlines from stdout
+        for v in reversed(stdout_split):
+            if v == '':
+                stdout_split.pop()
+            else:
+                break
```
tony, line 278:
```suggestion
        while stdout_split and stdout_split[-1] == "":
            stdout_split.pop()
```

@rockandska Thoughts on this?

rockandska: Cleaner and does the job :+1: 

Could squash it to preserve clean history if needed

tony:
@rockandska Since this _could_ be considered breaking behavior (-ish), I am going to release this under v0.15.0a0 for you to test

Standby !

tony:
@rockandska 

v0.15.0a0: [GitHub release](https://github.com/tmux-python/libtmux/releases/tag/v0.15.0a0), [git tag](https://github.com/tmux-python/libtmux/tree/v0.15.0a0), [PyPI package](https://pypi.org/project/libtmux/0.15.0a0/)

```console
pip install libtmux==0.15.0a0
```

tony:
Live in v0.15.0

PR #444, opened 2022-09-26
Allow empty window name

trankchung:


src/libtmux/session.py
```diff
@@ -253,6 +253,9 @@ def new_window(
         if window_name:
             window_args += ("-n%s" % window_name,)
 
+        if window_name == '':
+            window_args += ('-n', '')
```
tony, line 257: @trankchung Are you open to adding a test?

tony:
@trankchung I am going to bed soon - I may not get to this until tomorrow

src/libtmux/session.py
trankchung: @tony Yes I’ll add a test for this tomorrow.

tony: @trankchung Sounds good!

trankchung:
@tony Would you review please?

src/libtmux/session.py
tony, line 257:
```suggestion
        if window_name is not None and isinstance(window_name, str):
            window_args += ("-n%s" % window_name,)
```

```suggestion
        if window_name is not None:
            window_args += ("-n%s" % window_name,)
```

Would either of these accomplish the same?

tests/test_window.py
```diff
@@ -290,3 +290,11 @@ def test_select_layout_accepts_no_arg(server: Server, session: Session) -> None:
...
     window = session.new_window(window_name="test_window")
     window.select_layout()
+
+
+def test_empty_window_name(session: Session) -> None:
+    session.set_option("automatic-rename", "off")
+    window = session.new_window(window_name="''", attach=True)
+
+    assert window == session.attached_window
+    assert window.get("window_name") == "''"
```
tony, line 300: What's interesting is this test always passes.

If you remove from the changes from session.py and run:

```console
poetry run py.test tests/test_window.py::test_empty_window_name -vv -x -s
```

Do they still pass?

The implication is maybe we're looking at the wrong spot 🤔

src/libtmux/session.py
trankchung: I don’t think that would do it because a non-empty string would satisfy that condition.

tests/test_window.py
trankchung: How do you suggest to test this?

src/libtmux/session.py
tony: Both `''` and `'normal string'` would, no?

tests/test_window.py
tony: I will need to look at this during the weekend when I have more time, or tomorrow evening if I have time 

(i work during the day US time)

src/libtmux/session.py
trankchung: It should work

tests/test_window.py
trankchung: Yeah no rush

src/libtmux/session.py
tony: Would you take the above example as a compromise? 

I think the reason why the above works in one condition is it eliminates the `("-n%s" % window_name,)`. That's where the `"''"` is being lost.

What do you think?

tests/test_window.py
tony: If you haven't yet you can clone libvcs and tmuxp, then do `pip install --user -e ~/path/to/tmuxp`, `pip install --user -e ~/path/to/libtmux` and you should be able to test it by hand.

If you have `entr(1)` installed:

```console
make watch_test test='tests/test_window.py::test_empty_window_name -vv -x -s'
```

If no `entr(1)`:

```console
env PYTEST_ADDOPTS="-s -x -vvv tests/test_window.py::test_empty_window_name" poetry run make start
```

I actually don't know why the test passes always yet

trankchung: I read the code and it looks like `Window.new_window()` ran a command against the OS and then forgets about it. What I want to know is where does `Window.get()` gets the window_name from. I can't find `Window.get()` in the code. Do you mind pointing me to it? I can't find it in the inherited class neither.

I order to test this we need to be able to ask tmux itself for the window name and compare it against the expected value. Is there a way to ask tmux server for the current window name vs getting it from `Window.get()`?

tony: Don't sweat this too much. You can add a todo note.

If you want to wait until the weekend, I can provide more details, or we can merge this as is

trankchung: We can wait. No rush.

tony: @trankchung I can do it potentially. The API underneath the hood is an artifact of when I was new to python (it shows). To answer your question:

-   [TmuxMappingObject.__getattr__](https://github.com/tmux-python/libtmux/blob/v0.15.7/src/libtmux/common.py#L321-L328) (see [python.org](http://python.org/) docs for [Object.__getattr__(self, name)](https://docs.python.org/3/reference/datamodel.html#object.__getattr__))

-   [Window._info](https://github.com/tmux-python/libtmux/blob/v0.15.7/src/libtmux/window.py#L104-L124)

-   [Server._windows](https://github.com/tmux-python/libtmux/blob/v0.15.7/src/libtmux/server.py#L103)* <- *[Server._list_windows()](https://github.com/tmux-python/libtmux/blob/v0.15.7/src/libtmux/server.py#L215-L266)

This API is very old - I will be replacing the code around this to be more transparent
https://github.com/tmux-python/libtmux/pull/426

trankchung: @tony Okay sounds good. Let me know when it's merged and I'll use it for the test.

tony: @trankchung How about I merge yours as-is? (so the fix is there?)

tony: That #426 PR may not be merged for a while (not sure)

trankchung: > @trankchung How about I merge yours as-is? (so the fix is there?)

As long as you're okay with it.

tony: @trankchung I will take a look again in the AM tomorrow 👍 I've tested this by hand so I am feeling optimistic

tony: @trankchung I've created #445 to sort out why the test isn't working.  I suspect we're incorrectly filtering it out as a tmux format that was blank, in reality it's an empty string.

tony (approved):
Tested this by hand, this works well

Thank you @trankchung !

For the test, it's something on our side, I believe: #445. We're likely filtering out `''` as if the filter had no value.

I will create a new release now for libtmux and tmuxp so you can try

tony:
@trankchung [v0.15.8](https://pypi.org/project/libtmux/0.15.8/)

If you update libtmux by itself, tmuxp should work. You can verify libtmux version is 0.15.8 via `tmuxp -V`

(though its not pinned to 0.15.8 yet, doing that next)

tony:
@trankchung This fixed is pinned in tmuxp [v1.16.1](https://pypi.org/project/tmuxp/1.16.1/)

trankchung:
Thanks @tony. I'm a long time tmuxinator user but I now prefer tmuxp more.

tony:
@trankchung welcome to the fold

PR #452, opened 2022-11-01
fix(test_capture_pane): use of double `env`

zappolowski:
This makes the test fail (at least for me) as it still won't start bash but my login-shell (which is not bash).

### DIFF
--- a/tests/test_pane.py
+++ b/tests/test_pane.py
@@ -1,5 +1,6 @@
 """Tests for libtmux Pane object."""
 import logging
+import shutil
 
 from libtmux.session import Session
 
@@ -66,10 +67,13 @@ def test_set_width(session: Session) -> None:
 
 
 def test_capture_pane(session: Session) -> None:
+    env = shutil.which("env")
+    assert env is not None, "Cannot find usable `env` in PATH."
+
     session.new_window(
         attach=True,
         window_name="capture_pane",
-        window_shell='env -i PS1="$ " /usr/bin/env bash --norc --noprofile',
+        window_shell=f"{env} PS1='$ ' bash --norc --noprofile",
     )
     pane = session.attached_window.attached_pane
     assert pane is not None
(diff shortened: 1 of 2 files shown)


tests/test_pane.py
```diff
@@ -69,7 +69,7 @@ def test_capture_pane(session: Session) -> None:
     session.new_window(
         attach=True,
         window_name="capture_pane",
-        window_shell='env -i PS1="$ " /usr/bin/env bash --norc --noprofile',
+        window_shell="/usr/bin/env PS1='$ ' bash --norc --noprofile",
```
tony, line 72: Would [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which) be worth doing, Extra portability?

zappolowski: TBH I'm not sure how this would affect portability. I would highly assume that `bash` is always on `PATH` for all practical purposes.

zappolowski: I've never came across a system which didn't have `env` at `/usr/bin/env` (as this is used in quite a lot of shebangs at the moment and this would break then). I've added some lookup anyhow (will rebase if accepted).

tony: thank you.

tony: you're right in all your points

i wouldn't hold anything past unix-like container/cloud and raspberry pi /single board-distros breaking convention.

zappolowski: I've pushed a new version with the fixup commit fixed up :)

tests/test_pane.py
```diff
@@ -66,10 +67,13 @@ def test_set_width(session: Session) -> None:
...
 
 def test_capture_pane(session: Session) -> None:
+    env = shutil.which("env")
+    assert env is not None, "Cannot find usable `env` in PATH."
+
     session.new_window(
         attach=True,
         window_name="capture_pane",
-        window_shell='env -i PS1="$ " /usr/bin/env bash --norc --noprofile',
+        window_shell=f"{env} PS1='$ ' bash --norc --noprofile",
```
tony, line 76: But I was also saying over here that `bash` may not exist 😆  I will need to look at that later.

PR #455, opened 2022-11-05
test(test_capture_pane): Remove bashism(s)

### LINKED ISSUE #454: Remove BASH-isms from tests
re: https://github.com/tmux-python/libtmux/pull/452#discussion_r1014638444, https://github.com/tmux-python/tmuxp/pull/844

tony:
Fixes #454

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,8 +16,12 @@ $ pip install --user --upgrade --pre libtmux
 
 ### Tests
 
+_There will be more tests over the coming weeks and months to shore up
+flakiness across shells and environments._
+
 - Compatibility improvement for `test_capture_pane` and `env` (#452), credit:
   @zappolowski!
+- Remove more BASHisms from tests (#455)
 
 ## libtmux 0.15.9 (2022-10-30)
 
--- a/src/libtmux/common.py
+++ b/src/libtmux/common.py
@@ -219,7 +219,7 @@ class tmux_cmd:
 
     Equivalent to:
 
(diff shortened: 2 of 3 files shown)


tony:
@zappolowski How is this for you? (in re: https://github.com/tmux-python/tmuxp/pull/844)

Do any tests fail w/ libtmux?

zappolowski:
It's still the doctest from `src/libtmux/pane.py:159` which fails. That test also seems a bit flaky as the captured value either is empty `[]` or (as mentioned somewhere else) the default prompt of bash which is `PS1=\s-\v$ ` and thus `['sh-5.1$']`.

I still have no clue how this issue could be resolved, but as already stated its easy to ignore for humans - as long as the CI uses `dash` it won't fail.

tony:
I will take care of it - but I don't have an ETA yet (until I understand it better / reproduce it reliably)

tony:
@zappolowski This is an existential issue for libtmux / tmuxp

I think it's going to take some thinking.

Ideas: 

- sending a sentinel command made to figure out what the PS1 and shell is.
- building in support for polling `capture_pane()` (similar to playwright / selenium)
- finding "flake-free" / "rock-solid" loadouts for capturing pane text. Reliable enough to be used downstream in tests and across platforms

What we need:

- We need to know what `capture_pane()` and etc. is being used for.  Can that spare to use a no-frills shell if it gives them rock-solid results across platforms?

Issue #472, opened 2023-01-21
The parse of list-panes output breaks when the screen buffer contains pipe symbol


hongwen000:
When there are pipe symbols in the screen buffer, the buffer_sample field in the output of "tmux list-session" may contain pipe symbols. 

In the "fetch_objs" function of neo.py, the "obj_formatters" use pipe symbols as the delimiter between fields when parsing the output of "tmux list-session", causing an error in the parsing when pipe symbols are present in the buffer_sample field. 

This results in failure of all functions that rely on "fetch_objs" such as creating and listing sessions.

tony:
@hongwen: I followed up with you [here](https://github.com/tmux-python/libtmux/pull/473#issuecomment-1407426187), Can you try this PR or that environmental variable out and see if it allieves the issue?

tony:
@hongwen000 If you try [v0.21.0](https://pypi.org/project/libtmux/0.21.0/) (which includes #475), is anything different?

hongwen000:
@tony Yes, I have upgraded both tmuxp and libtmux, and now they are working well. Thanks for your following up.

tony:
@hongwen000 Excellent! Back in action now!

I will close this issue for now. Is #473 settled as well?

hongwen000:
@tony Yes. I have closed that pull request.

tony:
@hongwen000 Good to know this is fixed! I will close this for now. Let me know if this persists!

Issue #471, opened 2023-01-19
fetch_objs breaks when buffer_sample contains pipe symbol |


hongwen000:
The following is a sample output. In this output, the correct value of buffer_sample is "eval /home/aaa/miniconda3/bin/conda "shell.fish" "hook" $argv | source\n", however, fetch_objs recognizes the pipe symbol as a delimiter, resulting in an error in the entire parsing process.

```
|0|4294967295|4294967295|buffer445|eval /home/aaa/miniconda3/bin/conda "shell.fish" "hook" \$argv | source\n|72||||||||||||||||||/home/lixr/.tmux.conf|||||| |1|38|0|4844|50000|1|0|1|1|0|0|0|0|0|0|0|$19|0|1|default|65|zsh|/home/aaa||||default|66|%31|0|0|865720|269|||/home/aaa|8,16,24,32,40,48,56,64,72,80,88,96,104,112,120,128,136,144,152,160,168,176,184,192,200,208,216,224,232,240,248,256,264|0|/dev/pts/37|270|12830||65|0||||||1674096551||0||1674096551|||||$7||0119_1024_23_art_ci|/home/lixr/repo/androidtools|0|1|/tmp/tmux-1011/default|1672646611|1011|aaa|3.3a|1|0|1|1674096552|7|2|66|@17|0|6461,270x66,0,0,31|0|1|0119_1024_23_art_ci|0|zsh|||1|*|0|270|1|
```

PR #473, opened 2023-01-21
Fix tmux list-session parsing error by removing buffer_sample field from fetch_objs function

hongwen000:
See issue #472 and issue #471

### DIFF
--- a/src/libtmux/neo.py
+++ b/src/libtmux/neo.py
@@ -42,7 +42,8 @@ class Obj:
     # See QUIRK_TMUX_3_1_X_0001
     # buffer_created: t.Union[str, None] = None
     buffer_name: t.Union[str, None] = None
-    buffer_sample: t.Union[str, None] = None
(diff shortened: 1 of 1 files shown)


tony:
@hongwen000 Thank you for this contribution and the issues

There's another possibility where we can possibly keep the separator: We could change the separator from `|` to be something else, e.g. ␞ (see https://stackoverflow.com/a/29811033)

If you try `env LIBTMUX_TMUX_FORMAT_SEPARATOR=␞`, do you still get this problem? See also #475

hongwen000:
@tony Yes, this modification can also solve my problem. Changing the delimiter to a rare Unicode character is a nice idea. Although the this character existing in buffer_sample is theoretically possible, but it is almost unlikely. One "safe" solution is to call tmux list-session twice, to get buffer_sample and other fields separately. But this would increase additional complexity. In any case, I agree to change the delimiter to ␞ to solve this problem.

tony:
@hongwen000 I am going to release #475, come back, and see if it alleives the issue for you.

PR #505, opened 2023-10-23
Fixed __eq__ for windows.

m1guelperez:
```python
for window in session.list_windows():
            if window.name == "test_window":
                target_window = window
                break
        if target_window == None:
            target_window = session.new_window(
                attach=True, window_name="test_window"
            )
```
This code failed before at `if target_window == None` because comparison for windows was implemented like that:

```python
def __eq__(self, other: object) -> bool:
        assert isinstance(other, Window)
        return self.window_id == other.window_id
```

However, the assert seems a bit off since it is a way to strong assumption. Thus I changed it to:

```python
def __eq__(self, other: object) -> bool:
    if not isinstance(other, Window):
        return False
    return self.window_id == other.window_id
```

This code now properly returns `False` when we compare a window with `==` and the condition is indeed `False`.


JThoennes:
Actually, I would try to avoid negations with `not` and express the code like this:
```python
def __eq__(self, other: object) -> bool:
    if isinstance(other, Window):
        return self.window_id == other.window_id
    return False
```
Just my two cents.

m1guelperez:
> @m1guelperez Can you [sign the CLA](https://github.com/tmux-python/libtmux/pull/505#issuecomment-1826390707)?

Done :)

PR #506, opened 2023-11-18
Formatting: `black` -> `ruff format`

tony:
# Changes

## Development

- Convert from `black` to `ruff format` (https://docs.astral.sh/ruff/formatter/)

  This retains the formatting style of `black` while eliminating a dev dependency by using our existing rust-based `ruff` linter.

### DIFF
--- a/.vim/.vim/coc-settings.json
+++ b/.vim/.vim/coc-settings.json
@@ -1,0 +1,19 @@
+{
+  "[markdown][python]": {
+    "coc.preferences.formatOnSave": true
+  },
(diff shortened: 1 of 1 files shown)


piotr-dobrogost:
Any rationale for replacing black with ruff?

piotr-dobrogost:
Thanks for the answer.
Can't say anything about ruff as I haven't used it. I was asking just out of curiosity. I suggest trying to give short rationale for changes like this so that they don't appear "out of the blue".

Issue #530, opened 2024-03-01
No implicit exports from __init__.py


ssbarnea:
As https://github.com/tmux-python/libtmux/blob/master/src/libtmux/__init__.py we can see some imports that are not mentioned inside an `__all__` line.

This is is causing type check failures when running mypy in strict mode as it means that these module do not have an explicit exports. Example:

```
error: Module "libtmux" does not explicitly export attribute "Server"  [attr-defined]
```

PR #531, opened 2024-03-01
Add implicit exports into __init__.py

ssbarnea:
Fixes: #530


ssbarnea:
@tony Please note that this CLA checker is broken, see below. I even disabled my ad blocker and it still displays and empty form and no buttons.

PR #543, opened 2024-08-29
Fix docstring for `color` parameter

TravisDart:
This PR fixes an incorrect docstring.

For the `color` parameter, the API for libtmux actually expects either `256` or `88`. See this place in code: https://github.com/tmux-python/libtmux/blob/master/src/libtmux/server.py#L238


TravisDart:
Oops, deleted my fork too soon. Reopened here: https://github.com/tmux-python/libtmux/pull/544

Issue #556, opened 2025-01-30
server's socket_name and socket_path is empty


lazysegtree:
Server initialised without args have socket_name and socket_path as empty strings
```
>>> server = libtmux.Server()
>>> server
Server(socket_path=/tmp/tmux-1000/default)
>>> server.socket_path
>>> server.socket_name
>>> print(server)
Server(socket_path=/tmp/tmux-1000/default)
>>>
```

From the code it looks we dont initialise self.socket_path and self.socket_name, which we should, to default values. 
Is this by design, or is it something that we should fix ? If its to be fixed, I can raise a PR for it.

Some relavant code snippets

Server class `__init__`

```py
if socket_path is not None:
            self.socket_path = socket_path
        elif socket_name is not None:
            self.socket_name = socket_name

        tmux_tmpdir = pathlib.Path(os.getenv("TMUX_TMPDIR", "/tmp"))
        socket_name = self.socket_name or "default"
        if (
            tmux_tmpdir is not None
            and self.socket_path is None
            and self.socket_name is None
            and socket_name != "default"
        ):
            self.socket_path = str(tmux_tmpdir / f"tmux-{os.geteuid()}" / socket_name)

        if config_file:
            self.config_file = config_file

        if colors:
            self.colors = colors
```

String representation of Server object

```py
    def __repr__(self) -> str:
        """Representation of :class:`Server` object."""
        if self.socket_name is not None:
            return (
                f"{self.__class__.__name__}"
                f"(socket_name={getattr(self, 'socket_name', 'default')})"
            )
        if self.socket_path is not None:
            return f"{self.__class__.__name__}(socket_path={self.socket_path})"
        return f"{self.__class__.__name__}(socket_path=/tmp/tmux-1000/default)"
```

lazysegtree:
Also that fact that `/tmp/tmux-1000/default` is hardcoded, is wrong. In my mac, I see that socket path of `tmux` is different in my environment in MacOS . Its `/tmp/tmux-501/default`
```
➜  ~ [11:44:46] tmux list-session
no server running on /private/tmp/tmux-501/default
➜  ~ [11:44:50]
➜  ~ [11:48:40] ls -ltr /tmp
lrwxr-xr-x@ 1 root  wheel  11 Oct 15 16:52 /tmp -> private/tmp
➜  ~ [11:49:27]
```

So the `__repr__` method should be fixed.

PR #557, opened 2025-01-30
Fix hardcoded uid in __str__ method of Server class

lazysegtree:
See - https://github.com/tmux-python/libtmux/issues/556


lazysegtree:
> * You need to import the 'os' module at the top of the file to use os.geteuid()

Yes. Its already imported.

lazysegtree:
@tony Can you check this ?

lazysegtree:
Thanks @tony  . Anyways, many thanks for the library. I need to do a tmux automation, and I was going to write a wrapper myself, but I was glad when I found there was an existing library, and a well maintained one.

PR #562, opened 2025-02-07
Only use typing-extensions if necessary

ppentchev:
Hi,

Thanks for your continued work on libtmux!

Now here's a weird one. in a Debian package that I maintain, I wrote some tests that use libtmux and that are run by pytest. In the definition of the test environment for the Debian package, I listed libtmux and pytest as packages to be installed; neither of them brings in the typing-extensions library as a dependency, at least on recent versions of Python.

Now, when those tests were run, pytest detected the presence of `libtmux.pytest_plugin` and attempted to import it, even though my tests did not need it. Since the `libtmux.test` module has the `from typing_extensions import Self` line *outside* of a `TYPE_CHECKING` block, this led to the tests failing, since the running Python interpreter (correctly) could not find a `typing_extensions` module installed.

What do you think about these two commits? To be honest, the first one would be enough: moving the import inside a `TYPE_CHECKING` block resolves that particular problem. However, I got kind of carried away and I thought "why not make the import of `typing_extensions` conditional in the first place, since recent versions of Python do not really need that?" - hence the second commit that checks `sys.version_info` and imports `Self` and `TypeAlias` from `typing` instead of `typing_extensions` if possible.

I think that a dependency on `typing_extensions` should also be added to the `pyproject.toml` file - either unconditional as the code stands right now, or conditional on `python_version < "3.11"` or something. However, I have not yet used `uv` enough for developing modules (I am using it heavily to create test environments, and I am very happy with it), so I thought I'd leave that to people who are more familiar with that workflow.

Let me know if you think the first patch will be enough, and I will create a separate branch for it.

Thanks again, and keep up the great work!


shatakshiiii:
We are getting a similar issue on [ansible-navigator](https://github.com/ansible/ansible-navigator/actions/runs/13237446097/job/36945230770?pr=1916) where `python-3.13` tests fail if `typing-extensions` was not installed.

Thanks @ppentchev for coming up with the fix, a big +1 on this change!

shatakshiiii:
Thank you for merging this PR and rolling out a new release. I can confirm that the above mentioned failures are fixed and related tests are passing in ansible-navigator now with the latest version of libtmux.

TY @tony @ppentchev!!

PR #618, opened 2026-01-03
refactor(typing): Narrow t.Any to specific types across codebase

tony:
## Summary
- Replace `t.Any` with narrower types (`object`, `int | str`, etc.) throughout the codebase
- Add `TypedDict` definitions for structured keyword argument patterns
- Use function overloads for better type inference in temp helpers

### Core library
- **exc.py**: Narrow `VariableUnpackingError.variable` from `t.Any` to `object`
- **hooks.py**: Narrow `HookDict` values from `t.Any` to `int | str`
- **pane.py, session.py, window.py, server.py**: Narrow deprecated lookup methods (`get`, `__getitem__`, `where`, `find_where`) to use `object` instead of `t.Any`

### Test utilities
- **pytest_plugin.py**: Add `SessionParams` TypedDict for typed fixture parameters
- **test/temporary.py**: Add TypedDicts (`TempSessionParams`, `TempWindowParams`, etc.) and use overloads for `temp_session`/`temp_window` functions

### Tests
- Narrow test fixture types throughout (`ParameterSet`, specific union types)
- Tighten mock function signatures
- Replace generic `dict[str, t.Any]` with specific TypedDicts

## Test plan
- [x] All 872 tests pass
- [x] mypy passes with no issues
- [x] ruff linting clean

### DIFF
--- a/src/libtmux/exc.py
+++ b/src/libtmux/exc.py
@@ -351,7 +351,7 @@ class WaitTimeout(LibTmuxException):
 class VariableUnpackingError(LibTmuxException):
     """Error unpacking variable."""
 
-    def __init__(self, variable: t.Any | None = None, *args: object) -> None:
+    def __init__(self, variable: object | None = None, *args: object) -> None:
         return super().__init__(f"Unexpected variable: {variable!s}")
 
 
--- a/src/libtmux/hooks.py
+++ b/src/libtmux/hooks.py
@@ -50,7 +50,7 @@ from libtmux.options import handle_option_error
 if t.TYPE_CHECKING:
     from typing_extensions import Self
 
-HookDict = dict[str, t.Any]
+HookDict = dict[str, int | str]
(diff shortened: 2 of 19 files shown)


tony:
**Automated review finding.** Recorded here so it is not lost. Nothing has been pushed to this branch and no fix is applied — this is a note, not a change request.

### Conflicts with `#619` — needs a merge-order decision, not a patch

This PR's new `_temp_window` in `src/libtmux/test/temporary.py` reintroduces the `*args` forwarding to `Session.new_window` that `#619` removes as non-functional. Both PRs edit the same function against the same base blob, independently.

Consequence depends on order: if this merges after `#619`, it silently undoes that fix; if `#619` merges after this, the two conflict. Either way it is a maintainer decision about which change wins, not something to resolve mechanically — flagging it so the second merge is not a surprise.

Issue #624, opened 2026-02-05
Race condition in new_session() in some environments (e.g. PyInstaller + Python 3.13+ + Docker)


neubig:
`server.new_session()` fails with `TmuxObjectDoesNotExist` due to a seeming race condition in some settings.
Specifically, I noticed it in PyInstaller-bundled binaries running Python 3.13+ inside Docker containers.

## Root Cause

`new_session()` creates a session then immediately queries `list-sessions` to fetch session data. In certain environments, the session is not yet visible to `list-sessions`. Specifically, it seems that Python 3.13 and above encounters this issue, but Python 3.12 does not.

## Affected Environments

| Environment | Result |
|-------------|--------|
| Python 3.12 + PyInstaller + Docker | ✅ Pass |
| Python 3.13 + PyInstaller + Docker | ❌ **100% Fail** |
| Python 3.13 + PyInstaller + Native | ✅ Pass |

## Reproduction Script

```bash
#!/bin/bash
set -e
WORK_DIR=$(mktemp -d)
trap "sudo rm -rf $WORK_DIR 2>/dev/null || rm -rf $WORK_DIR" EXIT
cd "$WORK_DIR"

cat > test_libtmux.py << 'PYTEST'
import sys, uuid
import libtmux
from libtmux.exc import TmuxObjectDoesNotExist

server = libtmux.Server()
failures = 0
for i in range(10):
    try:
        session = server.new_session(session_name=f"test-{uuid.uuid4().hex[:8]}", x=80, y=24)
        session.kill()
    except TmuxObjectDoesNotExist:
        failures += 1
        print(f"  [{i+1}] FAILED")
print(f"Results: {10-failures}/10 passed")
sys.exit(1 if failures else 0)
PYTEST

docker run --rm -v "$WORK_DIR:/work" -w /work python:3.13-slim bash -c "
    apt-get update -qq && apt-get install -y -qq binutils tmux >/dev/null 2>&1
    pip install -q libtmux pyinstaller
    pyinstaller --onefile --clean --noconfirm --log-level ERROR test_libtmux.py 2>/dev/null
    ./dist/test_libtmux
"
```

PR #625, opened 2026-02-05
Fix race condition in new_session() by avoiding list-sessions query

neubig:
Fixes: https://github.com/tmux-python/libtmux/issues/624

Previously, new_session() would:
1. Run 'tmux new-session -P -F#{session_id}' to create session
2. Immediately run 'tmux list-sessions' to fetch full session data

This created a race condition in some environments where list-sessions might not see the newly created session yet, causing TmuxObjectDoesNotExist errors.

The fix:
- Expand the -F format string to include all Obj fields (except 'server')
- Parse the output directly into a Session object
- Eliminate the separate list-sessions query entirely

This is more efficient (one fewer subprocess call) and eliminates the race condition by making session creation atomic.

I tested this using the script in the original issue and it works in all three cases:

| Environment                        | Result   |
|------------------------------------|----------|
| Python 3.12 + PyInstaller + Docker | PASS ✓   |
| Python 3.13 + PyInstaller + Docker | PASS ✓   |
| Python 3.13 + PyInstaller + Native | PASS ✓   |

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -36,6 +36,26 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Bug fixes
+
+#### Fix race condition in new_session() (#625)
+
+Fixed {exc}`~libtmux.exc.TmuxObjectDoesNotExist` raised by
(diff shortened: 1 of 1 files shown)


neubig:
Great, thank you!

tony:
@neubig Thank you! This is live in v0.53.1:

- https://github.com/tmux-python/libtmux/releases/tag/v0.53.1
- https://pypi.org/project/libtmux/0.53.1/

Does this work okay for you downstream?

PR #626, opened 2026-02-05
Reproduce #624 race condition via real server crash

tony:
## Summary

- Add `test_new_session_race_condition` that reproduces #624 by injecting a real tmux server crash between `new-session` and `list-sessions`
- Monkeypatch controls only crash **timing** — all tmux commands, parsing, and error paths are exercised against real tmux
- Marked as `strict=True` xfail: will flip to XPASS when retry logic is added (the fix)

Related:
- #624
- #625

## How it works

1. Bumper session advances session IDs so `race_test` gets `$1`
2. `server.new_session("race_test")` runs real `tmux new-session` → returns `$1`
3. Patched `fetch_objs` intercepts the first `list-sessions` call: kills the server, starts a replacement (which gets `$0`)
4. Real `fetch_objs` runs `list-sessions` against the replacement server — finds only `$0`
5. `fetch_obj` searches for `$1`, doesn't find it → `TmuxObjectDoesNotExist`

## Test plan

- [x] `uv run pytest tests/test_server.py::test_new_session_race_condition -xvs` → XFAIL
- [x] `uv run ruff check tests/test_server.py` → clean
- [x] `uv run mypy tests/test_server.py` → clean
- [x] `uv run pytest tests/test_server.py -v` → 29 passed, 1 skipped, 1 xfailed

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -11,6 +11,7 @@ import typing as t
 
 import pytest
 
+from libtmux import exc
 from libtmux.server import Server
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


tony:
Closing — the race condition this PR reproduces no longer exists.

This PR (Mar 2026) demonstrates a server-crash race in `Server.new_session()` (issue #624) by patching `neo.fetch_objs(list_cmd="list-sessions")` to inject a server kill between `new-session` and the follow-up `list-sessions` query.

That race was fixed in #625 (released in v0.53.1, see CHANGES under `libtmux 0.53.1 (2026-02-18)` › "Fix race condition in new_session()"). The fix expands the `-F` format string in `new-session -P` to include all session fields and parses the output directly into a `Session` via `parse_output(session_stdout)` — eliminating the follow-up `list-sessions` query entirely.

As a result, the PR's test patch on `fetch_objs(list_cmd="list-sessions")` is unreachable. Confirmed by rebasing onto current master: `test_new_session_race_condition` is marked `xfail(strict=True)` and now triggers `XPASS(strict)` — pytest's signal that an "expected to fail" test is unexpectedly passing.

If you want to keep documenting this scenario for regression coverage, the test could be adapted to assert on the new code path's behavior under server crash (e.g. patching `Server.cmd` to fail mid-`new-session`). Otherwise this PR has served its purpose — the bug it documents has been fixed.

Confirmed during a 2026-04-26 batch rebase audit.

PR #641, opened 2026-03-08
fix: pre-existing issues in server.py and common.py (raise_if_dead, has_minimum_version)

tony:
## Summary

Three pre-existing issues surfaced during a fresh code review of PR #636. This PR addresses them as atomic commits against `master`.

- **P1** `common(fix[has_minimum_version])`: `has_minimum_version` was calling `get_version()` twice, spawning two subprocesses for no reason. Now cached in a local variable.
- **P2** `server(docs[raise_if_dead])`: `raise_if_dead` had no `Raises` section in its docstring. Both `TmuxCommandNotFound` (missing binary) and `CalledProcessError` (server not running) are now documented.
- **P3** `server(fix[raise_if_dead])`: `subprocess.check_call` in `raise_if_dead` inherited the parent process's stdout/stderr, leaking `list-sessions` output to the terminal. Now redirected to `subprocess.DEVNULL`.

## Test plan

- [x] `uv run ruff check . --fix --show-fixes` — clean
- [x] `uv run ruff format .` — no changes
- [x] `uv run mypy` — no issues (60 source files)
- [x] `uv run py.test -x -q` — 885 passed, 1 skipped

### DIFF
--- a/src/libtmux/server.py
+++ b/src/libtmux/server.py
@@ -330,7 +330,11 @@ class Server(
             cmd_args.insert(0, f"-f{self.config_file}")
 
         try:
-            subprocess.check_call([resolved, *cmd_args])
+            subprocess.check_call(
+                [resolved, *cmd_args],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
         except FileNotFoundError:
             raise exc.TmuxCommandNotFound from None
 


tony:
Flagging that this PR is **more current than its title suggests**, and that it silently fixes an issue filed four months after it.

The two things the title and body advertise have already landed on master independently — `has_minimum_version` caching (`ea561463`) and the `Raises` docstring (`85ace722`). What remains is a single commit touching one file, and its whole diff is:

```python
-            subprocess.check_call([resolved, *cmd_args])
+            subprocess.check_call(
+                [resolved, *cmd_args],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
```

That is exactly the fix #711 asks for: `Server.raise_if_dead()` shells out unredirected, so tmux's `no server running on …` / `error connecting to …` lands on the caller's fd 2 and cannot be intercepted with `contextlib.redirect_stderr`. `Server.is_alive()` goes through `self.cmd()` and is quiet, so the two liveness primitives behave inconsistently.

Worth retitling this PR to name that fix, and deciding one open question here rather than in a second PR: `DEVNULL` stops the leak but **discards** tmux's diagnostic entirely. Routing `raise_if_dead()` through `self.cmd()` would keep the stderr on the returned object where a caller can read it — but it would stop raising `subprocess.CalledProcessError`, which tests currently pin.

PR #653, opened 2026-03-29
tmux parity: add interactive command coverage and align flag semantics with tmux

tony:
## Summary

This PR expands libtmux's tmux command parity across `Server`, `Session`, `Window`, and `Pane`, with a focus on interactive and client-dependent commands. It also tightens several flag mappings and state-refresh behaviors so the high-level API matches tmux semantics more closely.

In addition to the new command coverage, this includes the three follow-up fixes from review:
- `Session.detach_client()` no longer forces `-s`, so targeted detach behaves like tmux.
- `Window.move_window()` now refreshes after successful moves, so returned objects are not stale after relative or cross-session moves.
- `ControlMode` now binds `client_name` to the spawned client by matching `client_pid`, which fixes multi-client races.

### New or expanded tmux command coverage
- Add `Server` support for commands such as `bind_key`, `unbind_key`, `list_keys`, `list_commands`, `start_server`, `lock_server`, `lock_client`, `refresh_client`, `suspend_client`, `server_access`, `run_shell`, `if_shell`, `source_file`, buffer commands, `confirm_before`, `command_prompt`, and `display_menu`.
- Add `Session.detach_client()` and window navigation helpers.
- Add `Window` support for `move_window` flags, `select_layout` flags, `last_pane`, `next_layout`, `previous_layout`, `rotate`, `swap`, `respawn`, `link`, and `unlink`.
- Expand `Pane` coverage for `display_popup`, `capture_pane`, `send_keys`, `select`, `copy_mode`, `clock_mode`, `choose_*`, `customize_mode`, `display_panes`, `find_window`, `paste_buffer`, `clear_history`, `pipe`, `join`, `break_pane`, `move`, `respawn`, and related flags.

### tmux semantics and compatibility fixes
- Correct several flag mappings and version gates to match tmux behavior.
- Treat `move-window -r` as standalone renumbering.
- Fix popup, prompt, paste-buffer, rotate, choose-tree, clear-history, capture-pane, and run-shell semantics where prior mappings diverged from tmux.
- Add version annotations and parity docs where new parameters were introduced on existing APIs.

### Test and tooling support
- Add `ControlMode` and pytest fixture support for commands that require a real attached client.
- Add extensive functional coverage for new command paths and regression cases.
- Add parity-analysis support files under `.claude/` and `skills/tmux-parity/` to help maintain command coverage against tmux.

## Testing

Ran:
- `uv run ruff check . --fix --show-fixes`
- `uv run ruff format .`
- `uv run mypy`
- `uv run py.test --reruns 0 -vvv`

Latest full run:
- `1067 passed, 1 skipped`

## Notes

- Diff vs `origin/master` is broad because this branch includes both the parity feature work and the review-driven correctness fixes on top.
- The attached-client test infrastructure is intentionally part of this PR because several new tmux commands cannot be exercised correctly without a real client.


tony:
### Code review

Found 1 issue:

1. `Server.command_prompt(bspace_exit=True)` emits `-e` with no version guard, while the sibling `literal` parameter on the same method does guard `-l` with `has_gte_version("3.6")` and a `warnings.warn` fallback. The docstring acknowledges that `-e` was added by upstream commit [`1e5f93b7`](https://github.com/tmux/tmux/commit/1e5f93b7) on 2026-01-14 and is "not in any tagged release at the time of writing", so any caller passing `bspace_exit=True` on currently-released tmux (3.2a–3.6a) will get an `unknown flag -e` error from tmux. The CI hides this because `tests/test_server.py::test_command_prompt_extra_flags[bspace_exit]` monkeypatches `server.cmd`, never invoking real tmux for the assertion.

https://github.com/tmux-python/libtmux/blob/095758ac6dd05f694add580b90c851f1e8d62bfa/src/libtmux/server.py#L1066-L1077

Suggested fix: mirror the `literal` pattern — guard the `-e` append on `has_gte_version` (e.g. `> 3.6` once a tag exists, or check `tmux master`) and emit a warning when the user requests `bspace_exit=True` on a version that lacks it.

Other items considered and dropped below the 80 confidence threshold: ControlMode `_write_fd` is not closed if `subprocess.Popen` raises before the registration loop ([control_mode.py L61–L92](https://github.com/tmux-python/libtmux/blob/095758ac6dd05f694add580b90c851f1e8d62bfa/src/libtmux/_internal/control_mode.py#L61-L92)) — real but error-path-only; and `Session.detach_client(all_clients=True, target_client=...)` enumerates `list-clients` then issues per-client detaches non-atomically — race window is microseconds and the docstring acknowledges it.

tony:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance against the 3 commits since the prior review (095758ac..1704a542).

PR #679, opened 2026-05-23
Subprocess(fix[encoding]): Enforce UTF-8 decoding for tmux output

### LINKED ISSUE #678: Enforce UTF-8 encoding
`tmux_cmd` opens the tmux subprocess with `text=True` but no explicit `encoding=`. Python defaults to the system locale. If the locale isn't UTF-8, the U+241E `FORMAT_SEPARATOR` gets mangled, so `parse_output()` can't split fields. All list accessors (`server.sessions`, etc.) return empty.
tmux has required UTF-8 since 2015 (`1b86f520`). Its output is always UTF-8.

The fix would just be:
```diff
 self.process = subprocess.Popen(
     cmd,
     stdout=subprocess.PIPE,
     stderr=subprocess.PIPE,
     text=True,
+    encoding="utf-8",
     errors="backslashreplace",
 )
```
in `src/libtmux/common.py:336`

tony:
## Summary

- Add regression test reproducing #678: `tmux_cmd` uses `subprocess.Popen(text=True)` without `encoding="utf-8"`, causing FORMAT_SEPARATOR (U+241E) corruption on non-UTF-8 locales
- Fix: add `encoding="utf-8"` to the `subprocess.Popen` call in `tmux_cmd.__init__`

### Commit sequence (hermetic proof)

| Commit | What | Test result |
|--------|------|-------------|
| `922a326a` | xfail regression test — monkeypatches `locale.getencoding` → `"latin-1"`, runs `list-sessions` through `tmux_cmd`, asserts FORMAT_SEPARATOR survives | **XFAIL** — separator corrupted to `â\x90\x9e`, assertion fails at the right place |
| `e05c8ca0` | Fix: `encoding="utf-8"` in `subprocess.Popen` — zero test changes | **XPASS(strict)** → CI fails, proving the fix resolves the reproduction |
| `c8279acf` | Remove xfail marker — zero code changes | **1257 passed** |

### Root cause

Commit `1a5e69a2` (2025-02-02) removed `console_to_str()` (which had an explicit UTF-8 fallback) and switched to bare `text=True`. Without `encoding="utf-8"`, CPython's `subprocess._text_encoding()` falls back to `locale.getencoding()`. On non-UTF-8 locales, the 3-byte UTF-8 sequence for `␞` (U+241E) is decoded as three latin-1 code points (`â\x90\x9e`), making `.split(FORMAT_SEPARATOR)` produce a single unsplit element. This cascades to `parse_output()` where `zip(..., strict=True)` raises `ValueError`, and list accessors like `server.sessions` return empty results.

tmux has mandated UTF-8 since 2015 — hardcoding `encoding="utf-8"` matches tmux's output contract.

Fixes: #678
Related: https://github.com/tmux-python/tmuxp/issues/1044

## Test plan

- [x] `uv run ruff check . --fix --show-fixes` — passes
- [x] `uv run ruff format .` — no changes
- [x] `uv run mypy` — no issues
- [x] `uv run pytest --reruns 0 -vvv` — 1257 passed, 2 skipped
- [x] `just build-docs` — builds successfully

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,17 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Fixes
+
+#### Subprocess encoding on non-UTF-8 locales (#679)
+
+{class}`~libtmux.common.tmux_cmd` and {class}`~libtmux._internal.control_mode.ControlMode`
+now pass `encoding="utf-8"` to `subprocess.Popen`, ensuring tmux output
+is decoded correctly regardless of the system locale. Previously, on
+non-UTF-8 locales, the {data}`~libtmux.formats.FORMAT_SEPARATOR` character
+(U+241E) was corrupted during decoding, causing list accessors
+({attr}`~libtmux.Server.sessions`, etc.) to return empty results.
+
 ## libtmux 0.57.1 (2026-05-18)
 
 Restores the "lenient-by-default" behavior for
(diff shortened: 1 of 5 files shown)


tony (commented):
### Code review

Issue #683, opened 2026-06-14
Typed command descriptors for native tmux command chains


tony:
## Summary

libtmux has prior tracking for this idea in #80, but that issue was closed as stale in 2018. The need is still current and sharper now: libtmux should have a typed way to describe the tmux command behind each wrapper method so callers can compose native tmux command sequences without dropping to raw `cmd(..., ";", ...)` calls.

This is also a downstream need for [`tmux-python/libtmux-mcp`](https://github.com/tmux-python/libtmux-mcp/issues/49). The MCP currently has tool-level batching for ordered operations, but that batching still sits above libtmux/tmux. A libtmux-native command descriptor and chain layer would let downstream tools opt into tmux's command queue semantics through typed Python APIs instead of hand-building argv or running several subprocesses.

Related context:

- Original closed tracker: #80
- MCP wrapper gaps: #670
- MCP batch-operation need: [`tmux-python/libtmux-mcp#49`](https://github.com/tmux-python/libtmux-mcp/issues/49)

## tmux behavior to model

tmux already has the native behavior this should expose:

- The manual defines semicolon-separated commands as a command sequence and says later commands do not run after a command in the sequence errors: [`tmux.1` at `3.6a`](https://github.com/tmux/tmux/blob/3.6a/tmux.1#L526-L582).
- tmux's argv parser receives already-tokenized arguments and splits commands on unescaped trailing semicolons: [`cmd-parse.y` at `3.6a`](https://github.com/tmux/tmux/blob/3.6a/cmd-parse.y#L1069-L1128).
- tmux turns the parsed command list into ordered command-queue items: [`cmd-queue.c` at `3.6a`](https://github.com/tmux/tmux/blob/3.6a/cmd-queue.c#L493-L535).

[...]

PR #684, opened 2026-06-14
Explore typed chainable tmux command APIs

tony:
## Summary

- Expands the isolated `tests/chainable_commands_experiment/` lab into an executable survey of typed tmux command-sequence API prototypes.
- Leaves shipped libtmux public APIs unchanged; no `src/libtmux` code is modified.
- Uses typed command values, typed target/ref demos, lazy query plans, async facades, and native semicolon command sequences to evaluate future libtmux batching APIs.
- Links to #683 and calls out the downstream libtmux-mcp need for one-dispatch native command sequences.

## Naming Outcome

The authoring APIs use batch/commands vocabulary rather than recorder vocabulary. `Recorder` reads like persisted history or event capture; these demos author future tmux command sequences.

```python
with api.CommandBatch() as commands:
    commands.session.new_window(window_name="work")
    commands.pane.split_window(horizontal=True, percentage=50)

sequence = commands.to_sequence()
```

## Shared Substrate

The lowest-level contract is an immutable command call or sequence. It renders to one native tmux argv sequence with `;` separator tokens, then dispatches once through a runner that looks like `Server.cmd()`.

```python
sequence = (
    CommandCall("new-window", ("-d", "-n", "work"))
    >> CommandCall("split-window", ("-h", "-p", 50))
)

assert sequence.argv() == (
    "new-window",
    "-d",
    "-n",
    "work",
    ";",
    "split-window",
    "-h",
    "-p",
    "50",
)
```

Tradeoff: this is a good internal IR, but too low-level to be the main user-facing API.

## Decorator API

Django-admin-style command metadata can be attached to normal typed functions.

```python
call = api.new_window(window_name="work", detach=True)

assert_type(call, CommandCall)
assert call.argv() == ("new-window", "-d", "-n", "work")
assert api.get_command_spec(api.new_window).scope == "session"
```

Tradeoff: excellent function-call ergonomics, but runtime-attached metadata needs helper protocols or stubs for static tools.

## Descriptor API

Descriptors bind concrete command objects and expose metadata on command-like attributes.

```python
commands = api.Commands()
call = commands.new_window(window_name="work", detach=True)

assert_type(commands.new_window, api.NewWindowCommand)
assert_type(call, CommandCall)
assert commands.new_window.spec.name == "new-window"
```

Tradeoff: strong metadata placement and completion, with more per-command boilerplate.

## Builder And Context APIs

Builder and batch-context APIs keep construction explicit and type-checker-friendly.

```python
expr = api.sequence(
    api.commands.new_window(window_name="work"),
) >> api.commands.split_window(horizontal=True, percentage=50)

assert_type(expr, CommandSequence)
```

```python
with api.CommandBatch() as batch:
    call = batch.new_window(window_name="work")
    batch.split_window(horizontal=True, percentage=50)

assert_type(call, CommandCall)
assert_type(batch.to_sequence(), CommandSequence)
```

Tradeoff: very testable, but separate from existing object methods unless a future integration exposes it from `Server`, `Session`, `Window`, or `Pane`.

## Auto-Batch And AST APIs

Transparent auto-batching can work for self-returning methods, but immediate-output methods need explicit deferred-result semantics.

```python
target = api.AutoBatchTarget()
returned = target.rename_window("work").select_layout("even-horizontal")

assert_type(returned, api.AutoBatchTarget)
assert target.to_sequence().argv() == (
    "rename-window",
    "work",
    ";",
    "select-layout",
    "even-horizontal",
)
```

The AST prototype keeps typed proxy calls and validates a restricted callable shape before executing it against a proxy that accumulates calls.

```python
def _script(proxy: api.AstCommandProxy) -> tuple[CommandCall, CommandCall]:
    return (
        proxy.new_window(window_name="work"),
        proxy.split_window(horizontal=True, percentage=50),
    )

sequence = api.from_callable(_script)
```

Tradeoff: both are useful research probes, but should stay research-only unless a later feature explicitly needs their constraints.

## Runnable And Orchestration APIs

Runnable composition gives symmetric one-off, batch, stream, and pipeline-style execution.

```python
pipeline = api.target_capture_call().then(api.run_command())

result = pipeline.invoke("%1", runner)
results = pipeline.batch(["%2", "%3"], runner)
streamed = list(pipeline.stream("%4", runner))
```

Prefect-style task submission maps deferred work to a native sequence.

```python
task = api.CommandTask(api.rename_window)
submitted = task.map((("@1", "editor"), ("@2", "logs")))
sequence = api.submitted_sequence(submitted)
```

Tradeoff: useful vocabulary for deferred work, but pipeline/concurrency naming may oversell what native tmux sequencing does.

## Query And Read APIs

QuerySet, async-query, lazy-plan, expression, selection, and document-query demos explore read/query surfaces.

```python
query = api.PaneQuery().filter(active=True).order_by("pane_index").limit(1)
rows = query.all(runner)
first = query.first(runner)
```

```python
query = (
    api.TmuxSelection.server()
    .sessions()
    .windows()
    .panes()
    .fields("pane_id", "pane_title")
)

payload = query.run(runner)
```

Tradeoff: strong for metadata/query APIs, less direct for mutating command sequences.

## Command Object API

The command-object demo makes each tmux command constructible, testable, composable, and runnable without dispatching immediately.

```python
command = api.PaneCmd.SplitWindow(
    target="%1",
    horizontal=True,
    percentage=50,
)

assert command.argv() == ("split-window", "-t", "%1", "-h", "-p", "50")
```

Command objects compose into native semicolon sequences and can be grouped through typed command namespaces.

```python
with api.CommandBatch() as commands:
    commands.session.new_window(window_name="work")
    commands.pane.split_window(horizontal=True, percentage=50)
    commands.window.select_layout(layout="even-horizontal")

assert commands.to_sequence().argv() == (
    "new-window",
    "-d",
    "-n",
    "work",
    ";",
    "split-window",
    "-h",
    "-p",
    "50",
    ";",
    "select-layout",
    "even-horizontal",
)
```

Tradeoff: this has the most boilerplate, but it gives the strongest combination of completion, static typing, fast argv tests, and controlled integration tests.

## Deferred Query Command Plan API

This demo combines typed targets, typed query rows, deferred `each()` command construction, pure snapshot compilation, and one-dispatch execution.

```python
plan = (
    api.panes()
    .filter(active=True)
    .order_by("pane_index")
    .each(lambda pane: [
        pane.cmd.send_keys("clear", enter=True),
        pane.cmd.resize_pane(height=20),
    ])
)

sequence = plan.to_sequence(snapshot)

assert sequence.argvs() == (
    ("send-keys", "-t", "%1", "clear", "Enter"),
    ("resize-pane", "-t", "%1", "-y", "20"),
    ("send-keys", "-t", "%2", "clear", "Enter"),
    ("resize-pane", "-t", "%2", "-y", "20"),
)
```

Rows are typed refs, not raw dictionaries, so row-bound command namespaces carry their target automatically.

```python
pane = snapshot.panes[0]

assert_type(pane.pane_id, api.PaneTarget)
assert pane.cmd.send_keys("clear", enter=True).argv() == (
    "send-keys",
    "-t",
    "%2",
    "clear",
    "Enter",
)
assert pane.window.select_layout("even-horizontal").argv() == (
    "select-layout",
    "-t",
    "@1",
    "even-horizontal",
)
```

`each()` is a side-effect command plan, while `map()` stays data-only and `flat_map()` is the explicit multi-command expansion API.

```python
titles = api.panes().filter(active=True).map(lambda pane: pane.title).all(snapshot)

plan = api.panes().flat_map(
    lambda pane: (
        pane.cmd.resize_pane(height=10),
        pane.window.select_layout("even-horizontal"),
    ),
)
```

Tradeoff: this is the strongest shape so far for target-safe deferred sequencing. It intentionally does not solve structured output commands such as `capture-pane`; those need a separate result strategy: per-item execution, delimiter-based parsing, or a typed output mode.

## Async Deferred Command Plan API

The asyncio demo keeps the same typed command-authoring surface as the sync deferred plan. Only tmux-state resolution and command dispatch become awaitable.

```python
async def resize_active_panes(runner: api.AsyncPlanRunner) -> None:
    plan = (
        api.panes()
        .filter(active=True)
        .order_by("pane_index")
        .each(lambda pane: pane.cmd.resize_pane(height=20))
    )

    sequence = await plan.to_sequence(runner)

    assert sequence.argvs() == (
        ("resize-pane", "-t", "%1", "-y", "20"),
        ("resize-pane", "-t", "%2", "-y", "20"),
    )

    await plan.run(runner)
```

Data queries are also awaitable, but remain separate from command construction.

```python
titles = await (
    api.panes()
    .filter(active=True)
    .order_by("pane_index")
    .map(lambda pane: pane.title)
    .all(runner)
)
```

Tradeoff: this proves the sync command IR can support an async transport without making command construction itself async. It does add facade boilerplate, so it only seems worth adopting if libtmux grows a real async runner or adapter layer.

## Recommendation

For a v1 internal design, explicit command objects plus deferred query command plans look strongest.

Command objects give the cleanest unit-test boundary: construct one command, assert `argv()`, compose it, and only run tmux in integration tests. Deferred query command plans add the missing target-safe ergonomic layer: resolve typed rows, bind commands from each row, inspect the native sequence with a fake snapshot, then execute once through tmux when safe. An async facade can reuse the same IR if libtmux needs awaitable snapshot and dispatch boundaries. Descriptors remain useful if command metadata should live on command-like attributes. AST and transparent auto-batching should remain research-only.

## Test Plan

- `uv run pytest tests/chainable_commands_experiment/test_async_deferred_plan_api.py` -> 4 passed
- `uv run pytest tests/chainable_commands_experiment` -> 65 passed
- `uv run mypy tests/chainable_commands_experiment` -> passed, 43 source files
- `uvx ty check tests/chainable_commands_experiment` -> passed
- `uv run ruff format . --check` -> 112 files already formatted
- `uv run ruff check .` -> passed
- `uv run mypy src tests` -> passed, 109 source files
- `rm -rf docs/_build && uv run pytest` -> 1323 passed, 2 skipped

Refs: #683

### DIFF
--- a/tests/chainable_commands_experiment/__init__.py
+++ b/tests/chainable_commands_experiment/__init__.py
@@ -1,0 +1,1 @@
+"""Typed command-chain API experiments for libtmux."""
--- a/tests/chainable_commands_experiment/ast_api.py
+++ b/tests/chainable_commands_experiment/ast_api.py
@@ -1,0 +1,98 @@
+"""AST-assisted command-chain experiment."""
+
+from __future__ import annotations
+
+import ast
+import inspect
+import textwrap
+from collections.abc import Callable
+
+from .shared import (
+    CommandCall,
+    CommandSequence,
+    new_window_call,
+    split_window_call,
+)
+
+
+class UnsupportedAstShape(ValueError):
+    """Raised when a callable uses syntax outside the AST prototype."""
+
+
+class AstCommandProxy:
+    """Typed proxy used by the AST experiment."""
+
+    def __init__(self) -> None:
+        """Initialize an empty proxy call list."""
+        self._calls: list[CommandCall] = []
+
+    def to_sequence(self) -> CommandSequence:
+        """Return accumulated calls as a command sequence."""
+        return CommandSequence(tuple(self._calls))
+
+    def new_window(
+        self,
+        *,
+        window_name: str | None = None,
+        detach: bool = True,
+    ) -> CommandCall:
+        """Add a ``new-window`` proxy call."""
+        call = new_window_call(window_name=window_name, detach=detach)
+        self._calls.append(call)
+        return call
+
+    def split_window(
+        self,
+        *,
+        horizontal: bool = False,
+        percentage: int | None = None,
+    ) -> CommandCall:
+        """Add a ``split-window`` proxy call."""
+        call = split_window_call(
+            horizontal=horizontal,
+            percentage=percentage,
+        )
+        self._calls.append(call)
+        return call
+
+
+def command_names_from_ast(
+    function: Callable[[AstCommandProxy], object],
+) -> tuple[str, ...]:
+    """Extract command method names from a restricted callable AST."""
+    function_def = _function_def_from_source(function)
+    if any(
+        isinstance(node, ast.For | ast.While | ast.If | ast.Match)
+        for node in ast.walk(function_def)
+    ):
+        msg = "control flow is outside this AST prototype"
+        raise UnsupportedAstShape(msg)
+
+    names = [
+        node.func.attr
+        for node in ast.walk(function_def)
+        if isinstance(node, ast.Call) and isinstance(node.func, ast.Attribute)
+    ]
+    if not names:
+        msg = "no proxy command calls found"
+        raise UnsupportedAstShape(msg)
+    return tuple(names)
+
+
+def from_callable(function: Callable[[AstCommandProxy], object]) -> CommandSequence:
+    """Validate a callable's AST, then execute it against a typed proxy."""
+    command_names_from_ast(function)
+    proxy = AstCommandProxy()
+    function(proxy)
+    return proxy.to_sequence()
+
+
+def _function_def_from_source(
+    function: Callable[[AstCommandProxy], object],
+) -> ast.FunctionDef:
+    source = textwrap.dedent(inspect.getsource(function))
+    module = ast.parse(source)
+    if len(module.body) != 1 or not isinstance(module.body[0], ast.FunctionDef):
+        msg = "expected a single function definition"
+        raise UnsupportedAstShape(msg)
+    return module.body[0]
--- a/tests/chainable_commands_experiment/async_deferred_plan_api.py
+++ b/tests/chainable_commands_experiment/async_deferred_plan_api.py
@@ -1,0 +1,172 @@
+"""Asyncio facade over deferred query-command plan experiments."""
+
+from __future__ import annotations
+
+import dataclasses
+import typing as t
+from dataclasses import dataclass
+
+from . import deferred_plan_api as sync_api
+from .shared import Arg, CommandCall, CommandResultLike
+
+MappedT = t.TypeVar("MappedT")
+ResultT = t.TypeVar("ResultT")
+
+NoCommandsResolved = sync_api.NoCommandsResolved
+
+
+class AsyncCommandRunner(t.Protocol):
+    """Object capable of asynchronously dispatching one tmux command argv."""
+
+    async def cmd(
+        self,
+        cmd: str,
+        *args: Arg,
+        target: str | int | None = None,
+    ) -> CommandResultLike:
+        """Dispatch one tmux command asynchronously."""
+        ...
+
+
+class AsyncSnapshotProvider(t.Protocol):
+    """Object that can asynchronously provide a pure tmux snapshot."""
+
+    async def snapshot(self) -> sync_api.TmuxSnapshot:
+        """Return a tmux snapshot asynchronously."""
+        ...
+
+
+class AsyncPlanRunner(
+    AsyncCommandRunner,
+    AsyncSnapshotProvider,
+    t.Protocol,
+):
+    """Runner that can resolve async queries and dispatch async commands."""
+
+
+AsyncSnapshotSource: t.TypeAlias = sync_api.TmuxSnapshot | AsyncSnapshotProvider
+
+
+@dataclass(frozen=True, slots=True)
+class PaneQuery:
+    """Async lazy pane query backed by the sync deferred query object."""
+
+    query: sync_api.PaneQuery
+
+    def filter(self, *, active: bool) -> PaneQuery:
+        """Return a query filtered by active state."""
+        return dataclasses.replace(self, query=self.query.filter(active=active))
+
+    def order_by(self, field: sync_api.OrderField) -> PaneQuery:
+        """Return a query ordered by a known pane field."""
+        return dataclasses.replace(self, query=self.query.order_by(field))
+
+    def limit(self, count: int) -> PaneQuery:
+        """Return a query capped to ``count`` rows."""
+        return dataclasses.replace(self, query=self.query.limit(count))
+
+    async def all(self, source: AsyncSnapshotSource) -> list[sync_api.PaneRef]:
+        """Evaluate the query against an async snapshot source."""
+        snapshot = await _resolve_snapshot(source)
+        return self.query.all(snapshot)
+
+    async def first(
+        self,
+        source: AsyncSnapshotSource,
+    ) -> sync_api.PaneRef | None:
+        """Evaluate the query and return its first row."""
+        snapshot = await _resolve_snapshot(source)
+        return self.query.first(snapshot)
+
+    def map(
+        self,
+        mapper: t.Callable[[sync_api.PaneRef], MappedT],
+    ) -> MappedPaneQuery[MappedT]:
+        """Return a data transformation query."""
+        return MappedPaneQuery(query=self, mapper=mapper)
+
+    def each(self, mapper: sync_api.CommandMapper) -> CommandPlan[None]:
+        """Return a deferred async side-effect command plan."""
+        return self.flat_map(mapper)
+
+    def flat_map(self, mapper: sync_api.CommandMapper) -> CommandPlan[None]:
+        """Return a deferred async multi-command side-effect plan."""
+        return CommandPlan(query=self, mapper=mapper)
+
+
+@dataclass(frozen=True, slots=True)
+class MappedPaneQuery(t.Generic[MappedT]):
+    """Async data-only query transformation over pane refs."""
+
+    query: PaneQuery
+    mapper: t.Callable[[sync_api.PaneRef], MappedT]
+
+    async def all(self, source: AsyncSnapshotSource) -> list[MappedT]:
+        """Evaluate the query and transform every row."""
+        return [self.mapper(row) for row in await self.query.all(source)]
+
+    async def first(self, source: AsyncSnapshotSource) -> MappedT | None:
+        """Evaluate the query and transform the first row."""
+        row = await self.query.first(source)
+        if row is None:
+            return None
+        return self.mapper(row)
+
+
+@dataclass(frozen=True, slots=True)
+class CommandSequence:
+    """Async wrapper around a resolved sync command sequence."""
+
+    sequence: sync_api.CommandSequence
+
+    @property
+    def calls(self) -> tuple[CommandCall, ...]:
+        """Return resolved command calls."""
+        return self.sequence.calls
+
+    def argvs(self) -> tuple[tuple[str, ...], ...]:
+        """Render each command independently."""
+        return self.sequence.argvs()
+
+    def argv(self) -> tuple[str, ...]:
+        """Render one native tmux semicolon command sequence."""
+        return self.sequence.argv()
+
+    async def run(self, runner: AsyncCommandRunner) -> CommandResultLike:
+        """Dispatch the sequence through one async runner call."""
+        argv = self.argv()
+        return await runner.cmd(argv[0], *argv[1:])
+
+
+@dataclass(frozen=True, slots=True)
+class CommandPlan(t.Generic[ResultT]):
+    """Async command plan that resolves a query into commands."""
+
+    query: PaneQuery
+    mapper: sync_api.CommandMapper
+
+    async def to_sequence(self, source: AsyncSnapshotSource) -> CommandSequence:
+        """Resolve the async query and compile mapped commands."""
+        snapshot = await _resolve_snapshot(source)
+        plan = self.query.query.flat_map(self.mapper)
+        return CommandSequence(plan.to_sequence(snapshot))
+
+    async def run(self: CommandPlan[None], runner: AsyncPlanRunner) -> None:
+        """Resolve, compile, and execute the plan through one async dispatch."""
+        try:
+            sequence = await self.to_sequence(runner)
+        except NoCommandsResolved:
+            return None
+        await sequence.run(runner)
+        return None
+
+
+def panes() -> PaneQuery:
+    """Start an async lazy pane query."""
+    return PaneQuery(sync_api.panes())
+
+
+async def _resolve_snapshot(source: AsyncSnapshotSource) -> sync_api.TmuxSnapshot:
+    if isinstance(source, sync_api.TmuxSnapshot):
+        return source
+    return await source.snapshot()
--- a/tests/chainable_commands_experiment/async_query_api.py
+++ b/tests/chainable_commands_experiment/async_query_api.py
@@ -1,0 +1,57 @@
+"""Piccolo-style async query experiment."""
+
+from __future__ import annotations
+
+import dataclasses
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True, slots=True)
+class AsyncPaneRow:
+    """Small async pane row."""
+
+    pane_id: str
+    active: bool
+
+
+@dataclass(frozen=True, slots=True)
+class AsyncPaneRunner:
+    """Async runner returning fixed pane rows."""
+
+    rows: tuple[AsyncPaneRow, ...]
+
+    async def list_panes(self) -> list[AsyncPaneRow]:
+        """Return pane rows asynchronously."""
+        return list(self.rows)
+
+
+@dataclass(frozen=True, slots=True)
+class AsyncPaneQuery:
+    """Immutable async pane query."""
+
+    active_filter: bool | None = None
+    limit_count: int | None = None
+
+    def where(self, *, active: bool) -> AsyncPaneQuery:
+        """Return a query filtered by active state."""
+        return dataclasses.replace(self, active_filter=active)
+
+    def limit(self, count: int) -> AsyncPaneQuery:
+        """Return a query capped to ``count`` rows."""
+        return dataclasses.replace(self, limit_count=count)
+
+    async def all(self, runner: AsyncPaneRunner) -> list[AsyncPaneRow]:
+        """Evaluate the async query."""
+        rows = await runner.list_panes()
+        if self.active_filter is not None:
+            rows = [row for row in rows if row.active is self.active_filter]
+        if self.limit_count is not None:
+            rows = rows[: self.limit_count]
+        return rows
+
+    async def first(self, runner: AsyncPaneRunner) -> AsyncPaneRow | None:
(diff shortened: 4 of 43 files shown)


tony:
## Feedback from driving tmux through an MCP layer (libtmux-mcp)

Some real-world motivation for this exploration, from the perspective of an automation/agent layer that drives tmux through an MCP server built on libtmux.

### The pain today

Building a layout programmatically — say a right-hand column of four evenly stacked panes — currently takes four ordered tool calls, each splitting the pane id returned by the previous one:

```
split %1 right  -> %2
split %2 below  -> %3
split %3 below  -> %4
split %4 below  -> %5
```

Each call is a separate libtmux dispatch → separate `tmux` process → separate command queue. The new pane id doesn't exist until the call returns, so I can't express the whole layout up front; I'm forced into a dependency chain of round-trips, one per pane.

On the CLI the same thing is a single invocation:

```sh
tmux split-window -h \; split-window \; split-window \; split-window
```

That works because every command in one invocation shares tmux's `current` target state, and `split-window` advances it to the newly created pane — unless `-d` is set:

https://github.com/tmux/tmux/blob/3f651d9/cmd-split-window.c#L176-L177

libtmux's `split(attach=False)` always passes `-d`, so that cursor never advances, and because each MCP call is its own process the shared queue state is gone at the boundary anyway. Net effect: there's no way to fold a layout build into one atomic tmux invocation today.

### The pattern that would be the unlock

Exactly what this PR is prototyping: an ordered sequence of typed commands that compiles to one native `\;` invocation, with targets that can be deferred and resolved rather than known up front. The substrate is already the right shape — `CommandSequence.argv()` semicolon-joins the calls and `.run()` dispatches through a single `runner.cmd(...)`:

https://github.com/tmux-python/libtmux/blob/228ff0b/tests/chainable_commands_experiment/shared.py

Collapsing N round-trips into one invocation is the headline win; recovering tmux's own cursor semantics inside that single queue is the bonus.

### Deferred plan

`deferred_plan_api` is the variant that maps most directly onto what an automation layer needs:

https://github.com/tmux-python/libtmux/blob/228ff0b/tests/chainable_commands_experiment/deferred_plan_api.py

- Typed, target-safe targets (`PaneTarget` / `WindowTarget` / `SessionTarget`) so `select-layout` can only bind a window and `send-keys` only a pane — mistakes that would otherwise surface at runtime.
- A lazy query over a snapshot, mapped to commands, compiled once and dispatched once (`CommandPlan.run` → `to_sequence(snapshot)` → `sequence.run(runner)`), with `NoCommandsResolved` degrading to a no-op.

The big deal: I never thread freshly created ids through arguments. I query live state, map each row to commands, and let the plan resolve and dispatch atomically.

The companion `auto_batch_api` is worth keeping alongside it — it encodes the detection rule for "what may fold into a chain": a call is chain-safe iff its output isn't consumed mid-chain. `DeferredCommandResult` raises if you read stdout before dispatch, and `show_option` is the deliberate counterexample of a call that can't batch because it needs output immediately:

https://github.com/tmux-python/libtmux/blob/228ff0b/tests/chainable_commands_experiment/auto_batch_api.py

Combined with the `CommandSpec.chainable` flag, that's both the static and the dynamic half of deciding what a batch compiler may merge.

### What an async deferred plan adds on top

The async facade is more than a coat of paint for anything hosting this inside a server:

https://github.com/tmux-python/libtmux/blob/228ff0b/tests/chainable_commands_experiment/async_deferred_plan_api.py

- The runner protocol becomes `async def cmd(...)` and snapshot acquisition becomes awaitable (`AsyncSnapshotProvider`), matching how an async server and async libtmux dispatch actually run — dispatching a layout no longer blocks the event loop or other in-flight calls.
- Snapshot reads can overlap with other awaits, and independent plans (e.g. several windows) can be resolved and dispatched concurrently.
- It stays a thin facade over the sync engine (it reuses `to_sequence`), so the single-native-sequence guarantee is preserved — one plan still compiles to exactly one `tmux ... \; ...` dispatch.

So: the sync deferred plan removes the round-trip and id-threading problem; the async deferred plan makes that same guarantee safe to use under a concurrent server without giving up the one-invocation-per-plan property.

One prerequisite worth calling out for the split-chain case specifically: an `attach` opt-out (so a sequence can intentionally let the `current` cursor walk into freshly created panes) interacts with the `-d` behavior linked above. The snapshot/deferred-target approach sidesteps needing that for target resolution, which is part of why it's the more attractive path.

tony:
## Review: the pattern this lab should converge on

I went through the whole `tests/chainable_commands_experiment/` lab — every prototype and its tests — against five axes: developer experience, typing, testability, maintainability, and expressiveness. This writes up the tradeoffs of all of them and names the single pattern they point at, building on the libtmux-mcp motivation already in this thread.

### The pattern this points at

> **An immutable argv command-sequence IR, with a typed, target-safe query → command-plan layer as the authoring surface, served through an async facade that preserves "one plan = one native `\;` dispatch", and governed by a chainability contract that decides what may fold into a single invocation. Mutations fold into one dispatch; reads that must return output stay individual typed calls.**

Concretely that is `deferred_plan_api` sitting on the `shared.py` IR, with `async_deferred_plan` as the hosted-server front door and the chainability rule from `auto_batch_api` deciding what merges. Everything else is either a piece of that pattern, a different ergonomic wrapper over the same IR, or a more ambitious shape that costs more than native tmux sequencing can repay.

### Why this is the unlock (the MCP case in this thread)

The motivating pain is real: building a four-pane column today is four ordered round-trips, because each new pane id doesn't exist until the previous call returns — every call is its own `tmux` process with its own command queue. On the CLI it's one invocation (`split-window -h \; split-window \; …`) because the commands share tmux's `current` target state. Two facts make the one-invocation path the prize:

- The substrate is already the right shape: `CommandSequence.argv()` semicolon-joins calls and `.run()` dispatches through a single `runner.cmd(...)` (`shared.py`). Collapsing N round-trips into one invocation is the headline win.
- There are **two ways to target panes inside one invocation**, and they cover different cases:
  - *Operate on objects that already exist* (resize every active pane, send-keys to matching panes): **deferred snapshot targets** — `deferred_plan_api` solves this completely and target-safely.
  - *Create and target within the same invocation* (the layout build): the new ids don't exist at authoring time, so you lean on tmux advancing its `current` cursor command-to-command. That requires **not** forcing `-d` — and today `Pane.split(attach=False)` always appends `-d`, so the cursor never advances. An `attach`/`-d` opt-out is the prerequisite for that specific case; the snapshot approach sidesteps it for everything else.

### The idea doing the heavy lifting: typed targets

The differentiator across the whole lab is the **typed-target row** in `deferred_plan_api`. Rows carry `pane_id: PaneTarget`, `window_id: WindowTarget`, `session_id: SessionTarget`, and expose `.cmd` / `.window` namespaces already bound to the right target — so `select-layout` can only bind a window and `send-keys` only a pane, and the mapper never touches a raw string. Every other prototype takes `target: str | int | None` positionally, which is easy to get wrong. Paired with **pure-snapshot compilation** (`to_sequence(snapshot)` is a pure function), the entire query → argv path is testable in memory with a fake snapshot — no tmux, sub-millisecond — with live tmux only at `.run()`. That is the strongest testability story here.

### What the async facade actually buys

`async_deferred_plan` is more than a coat of paint for anything hosting this in a server: the runner becomes `async def cmd(...)` and snapshot acquisition becomes awaitable, so dispatching a layout no longer blocks the event loop, snapshot reads overlap with other awaits, and independent plans (e.g. several windows) resolve and dispatch concurrently. Crucially it stays a thin facade over the sync engine — it reuses the sync `to_sequence` — so the one-native-sequence-per-plan guarantee is preserved. If the host is async (an MCP server is), this isn't optional polish; it's the front door.

### The chainability contract (what `auto_batch_api` is really for)

The valuable thing in `auto_batch_api` isn't the transparent-proxy ergonomics — it's the **rule it encodes**: a call may fold into a chain iff its output isn't consumed mid-chain. The dynamic half is demonstrated by `DeferredCommandResult`, whose `.stdout`/`.stderr`/`.returncode` raise `DeferredOutputUnavailable` until the chain runs, with `show_option` as the deliberate counterexample — a command that can't batch because it needs output immediately.

Worth stating precisely: the **static** half — a `CommandSpec.chainable` flag — exists in the IR but is **not yet wired** to that dynamic predicate; `auto_batch_api` doesn't reference it. Connecting the two (a static `chainable` declaration plus the runtime "output-not-consumed" guard) is exactly the contract a batch compiler needs to decide what may merge. That's the part to carry forward, not the proxy UX.

### The gap that sets the scope line

The hard limit is **structured per-command output** (`capture-pane`, `show-option`). One semicolon invocation returns one merged result — tmux gives no per-command demultiplexing. The lab confronts this honestly but doesn't solve it (`CapturePane` constructs but `run()` is flat; `deferred_plan` rejects data-returning mappers; `auto_batch` raises). The clean resolution is to keep the plan/sequence layer **mutation-only**: the ORM already does single-command read-back well (`split-window -P -F '#{pane_id}'`), and per-command demux inside a sequence would need fragile sentinel injection and drift toward transaction semantics this work explicitly treats as a non-goal.

### Tradeoffs of every prototype

| Prototype | Strength | Tradeoff / cost | Where it lands |
|---|---|---|---|
| `shared.py` | Tiny, stable argv IR; `;`-escaping; one-dispatch run; pure-testable | Too low-level to be the user-facing surface alone | **Foundation** |
| `deferred_plan_api` | Typed-target rows, pure-snapshot compile, `each`/`map`/`flat_map`, one dispatch | Mutation-only by design; doesn't address structured output | **The target shape** |
| `async_deferred_plan_api` | Awaitable runner + snapshot; concurrent plans; thin facade reusing sync `to_sequence` | Facade boilerplate | **Part of the target** when the host is async |
| `auto_batch_api` | Encodes the chainability predicate (`DeferredOutputUnavailable`; `show_option` counterexample) | Transparent-proxy UX is a footgun; static `chainable` flag not wired in | **Keep the predicate**, not the proxy |
| `command_object_api` | Concrete frozen-dataclass commands; `assert_type` precision; keeps live + stop-on-error tests | Most boilerplate; command declared as dataclass *and* forwarding namespace method; output is just a `CommandCall` | Largely redundant with the IR; add scoped values only where a typed `server.`/`pane.` namespace earns it |
| `context_api` | Clean `with`-block batch; construction fully separable from dispatch (no `.run()`) | One typed method per command — cheap to start, not to scale | Fine as a thin façade over command values |
| `builder_api` | Smallest explicit builder; type-checker-friendly | Same per-command method cost at scale | Fine as a thin façade |
| `decorator_api` | Best plain-call ergonomics; reads like a normal typed function | Metadata attached at runtime is invisible to static tools; needs stubs | Nice ergonomics, weak static-metadata story |
| `descriptor_api` | Cleanest place to hang metadata on command-like attributes | Per-command `@overload __get__` machinery is heavy | The issue's nominal framing; justified only if attribute-level metadata discovery (tool codegen) is proven necessary |
| `queryset_api` | Familiar ORM read surface (`filter/order_by/limit`) | Read-only; doesn't sequence mutations | Read side, not the command core |
| `async_query_api` | Awaitable read facade | One narrow idea | Read-side research |
| `selection_api` | Expressive `server().sessions().windows().panes().fields(...)` | Stringly-typed scope/field names | Read-side research |
| `expression_api` | Q-object-style composable predicates | An expression algebra most callers won't need | Only if filter composition gets complex |
| `document_query_api` | Metadata/document-style filtering | Narrow; composes with nothing | Research probe |
| `statement_api` | Compact statement surface | Too stringly-typed | Research probe |
| `lazy_plan_api` | Polars/lazy vocabulary | `optimize()` is a no-op; graph framing over a flat command list | Research only |
| `dag_api` | Dependency/topo-sort framing | Implies a dependency graph tmux sequencing lacks | Research only |
| `runnable_api` | Symmetric `invoke`/`batch`/`stream`/`pipeline` | `stream`/`pipeline` oversell one ordered dispatch | Research only |
| `orchestration_api` | Prefect-style `task.map(...)` → sequence | Concurrency/workflow language native sequencing doesn't provide | Research only |
| `ast_api` | Typed proxy + restricted-callable validation | Relies on `inspect.getsource`; can't handle lambdas/REPL; statically opaque | Research only |
| `generated_client_api` | Generated namespace from specs | Only pays off if a real codegen pipeline exists | Depends on a generator story not here |

### Breakdown by group

**The core (detailed above).** `shared.py` + `deferred_plan_api` + `async_deferred_plan` + the chainability predicate from `auto_batch_api` are the pattern; see the four sections above. Everything below is measured against them.

**Hand-authoring façades over the IR — `command_object_api`, `context_api`, `builder_api`, `decorator_api`.** All four let you author a fixed sequence without a query; they differ only in ergonomics. `command_object_api` gives the most precise types (concrete frozen dataclasses, `assert_type`) and the best test story (pure argv plus live stop-on-error), but the most boilerplate — each command is a dataclass *and* a forwarding namespace method, and its output is still just a `CommandCall`, so it's largely redundant with composing the IR directly (`CommandCall(...) >> CommandCall(...)`). `context_api` and `builder_api` are smaller and keep construction cleanly separate from dispatch, but still cost one typed method per command — cheap to start, not cheap to scale. `decorator_api` has the nicest call site (reads like a normal typed function) but attaches metadata at runtime where static tools can't see it. Net: adopt at most one, as a thin façade over command values, and only once callers want hand-authoring beyond raw `>>`.

**Metadata placement — `descriptor_api`.** This is the issue's nominal Django-`@admin.display` framing and the only demo that hangs metadata on a still-callable attribute. The cost is the per-command `@overload __get__` machinery, which is heavy and doctest-expensive. The lighter alternative already in the lab is the `CommandSpec` `ClassVar` on command values, which carries the same metadata with no extra machinery. Reach for descriptors only if attribute-level discovery (e.g. downstream tool codegen) is proven necessary.

**Read / query surfaces — `queryset_api`, `async_query_api`, `selection_api`, `expression_api`, `document_query_api`, `statement_api`.** A cluster exploring the read side: a familiar ORM `filter/order_by/limit`, awaitable reads, traversal (`server().sessions().windows().panes().fields(...)`), Q-object predicates, and metadata/statement filters. They're useful for typed reads, but they don't sequence mutations and their typing varies (`selection`/`statement` lean stringly-typed). They complement the snapshot the plan layer already reads rather than competing with the command core — and the plan's own `map()` over a snapshot covers the common "read into typed rows" need.

**Over-ambitious shapes (research only) — `lazy_plan_api`, `dag_api`, `runnable_api`, `orchestration_api`, `ast_api`, `generated_client_api`.** These impose graph (`dag`; `lazy_plan.optimize()` is a no-op), concurrency/workflow (`runnable`'s stream/pipeline, `orchestration`'s Prefect-style `task.map`), source-introspection (`ast` via `inspect.getsource` — can't handle lambdas or REPL-defined functions), or codegen (`generated_client`, contingent on a generator that isn't here) vocabularies. A single ordered semicolon dispatch has no dependencies, concurrency, or streaming to model, so these oversell what tmux provides. Keep them as probes; don't ship.

### A maintainability lever worth surfacing

Every module in the lab currently carries zero doctests and passes CI only by living under `tests/`. The test config enables `--doctest-docutils-modules` and includes `src/libtmux` in `testpaths`, so the moment any shape moves into `src/`, the project's "every public method has a working doctest, no skips" rule applies to every method. Doctest burden scales with public surface area — the strongest practical argument for leading with the bounded-surface query-plan layer and resisting a wide command catalog (and its duplicated namespace methods) before callers need it.

### Net

Carry forward `shared.py` as the IR, `deferred_plan_api` as the authoring surface, `async_deferred_plan` as the async front door, and the chainability predicate from `auto_batch_api` (wired to a static `CommandSpec.chainable` flag) as the merge-safety contract — plus an `attach`/`-d` opt-out for the create-within-one-invocation layout case. Keep AST, DAG, orchestration, streaming, lazy-plan, and the standalone command-object/descriptor surfaces as research until a caller proves each one earns its surface.

PR #685, opened 2026-06-14
Add experimental typed `chain` API for tmux command sequences

tony:
## Summary

- **Add** `libtmux._experimental.chain`, an experimental typed API for composing an ordered set of tmux commands that runs as **one** native `tmux ... \; ...` invocation instead of one subprocess per command. Promotes the converged design from the #684 research survey into real, documented, typed modules. Refs #683.
- **Five layers over one substrate** — an argv intermediate representation, a target-safe deferred query/expression layer, an async facade, a live-tmux connection layer (sync + async), and a chainability contract — all compiling down to a single `ir.CommandChain` and dispatching once.
- **Forward references resolve to not-yet-created objects.** A pane, window, or session reference can be *forward-declared* — split a pane, then address the pane that split will create before it exists. One object type carries both states (concrete row with typed metadata, or a pending placeholder that fails closed until tmux assigns the real id). A `ForwardPlan` resolves independent handles across all three scopes in the minimum number of dispatches, and folds a single pane handle back into one invocation via tmux's marked-pane register.
- **Chainability is enforced, not just declared.** The compile path rejects a command whose output would be consumed mid-chain (`show-option`, `capture-pane`), and a deferred result handle won't hand back output until the chain has run.
- **A typed escape hatch** lets callers issue any tmux command bound to a pane, window, or session target, so the layer isn't limited to its first-class builders.
- **Bundled** a small, non-experimental targeting fix to `Session.kill_window` / `Pane.break_pane`, now with regression tests.
- **Document** everything under a new `docs/experiment/` tree with runnable doctest examples; the package is explicitly outside the versioning policy and not re-exported from top-level `libtmux`.

### Experimental package — `src/libtmux/_experimental/chain/`

| Module | Role |
|---|---|
| **`ir.py`** | Immutable argv IR: `CommandCall`, `CommandChain` (`argv`/`argvs`/`>>`/`run`), `CommandSpec`, `SlotRef` (an unresolved "id of forward slot N" target), runner protocols, `;`-escaping. Rejects an empty-string target at construction; logs each one-shot dispatch (`tmux_cmd`/`tmux_subcommand`/`tmux_exit_code`). |
| **`plan.py`** | Typed targets, command values, dual-purpose `PaneRef`/`WindowRef`/`SessionRef` rows (concrete **or** forward) with bound `.cmd`/`.window`/`.session` namespaces (each carrying a `raw(name, *args)` escape hatch) and forward creation verbs (`split`, `new_window`, `break_pane`); pending metadata fails closed with `ForwardDataUnavailable`. Lazy `PaneQuery` with `.map` (data-only rows) and `.commands` (one-or-more commands per row), `CommandPlan` (pure `to_chain(snapshot)`, one-dispatch `run`, and `run_deferred` returning resolved per-command handles). |
| **`_resolve.py`** | Multi-dispatch resolution for independent forward handles: `ForwardPlan`/`ForwardHandle` (hand out independent handles across pane/window/session scopes), `Resolved` (slot→concrete-id bindings). A **sans-I/O** generator core yields a request and resumes via `.send(result)` — the same trampoline asyncio itself uses — driven identically by a sync (`run_resolving`) or async (`run_resolving_async`) driver. A plan-shape analyzer folds a lone pane handle into one `{marked}` invocation and otherwise resolves over N dispatches. |
| **`_async.py`** | Async facade mirroring the query/dispatch API; reuses the sync compile path so one expression still yields one `CommandChain`. Provides async `run`/`run_deferred`. |
| **`_connection.py`** | Live-tmux bridge: `snapshot_from_session` (skips panes missing a window/session id), `SessionPlanExecutor`, `AsyncSessionPlanExecutor`. |
| **`chain.py`** | Chainability contract: `COMMAND_SPECS` / `is_chainable` (static, wired into `to_chain` → `ChainabilityError`) plus `DeferredCommandResult` / `DeferredOutputUnavailable` (dynamic — a two-state handle that resolves to the chain's merged result after dispatch). |

### Forward references and multi-dispatch resolution

- **Dual-purpose refs (`plan.py`).** `PaneRef`/`WindowRef`/`SessionRef` each model one type in two states: a concrete row (typed `pane_index`/`active`/`title`) built via `.concrete(...)`, or a forward placeholder whose creation verbs (`split`, `new_window`, `break_pane`, top-level `new_session`) return another ref. The propagated parent target keeps a forward child addressable (e.g. a forward pane still knows the concrete window it was split in). Pending metadata access raises `ForwardDataUnavailable` rather than silently returning `None` (pending-attribute pattern).
- **Single sans-I/O core (`_resolve.py`).** Resolution is a pure generator that yields a `Dispatch`/`SnapshotRequest` and resumes via `.send(result)`; two ~10-line drivers (sync `runner.cmd`, `await runner.cmd`) share it verbatim, so the N-dispatch logic is never duplicated and the generator suspends at a `yield` between dispatches (never blocking the loop).
- **Independent handles → N dispatches.** A native `\;` sequence can only address the active (or single marked) object with no `-t` — a fixed argv token can't carry a freshly-created id, which escapes only as `-P -F` stdout. So independent handles run one creation per dispatch (`-P -F '#{pane_id}'` to capture the real id), then fold downstream commands into one trailing `\;` chain with the captured ids substituted. The builder spans all three scopes: `new_session` → `new_window` (targets the session via the captured `$N:`) → `split`.
- **Single pane handle → one `{marked}` dispatch.** A plan-shape analyzer detects the one case that folds back to a single invocation: a lone pane creation. It captures the id, marks the new (active) pane with `select-pane -m`, addresses it through tmux's server-wide `{marked}` register for every downstream command (verified against the tmux source to resolve within the same `\;` sequence), and clears the mark with `select-pane -M`. Two or more independent handles, or a detached session creation, stay on the N-dispatch path automatically.

### Production fix — `src/libtmux/pane.py`, `src/libtmux/session.py`

- **Fix** `Session.kill_window` and `Pane.break_pane` to scope a bare window target to the owning session, so it no longer resolves against the server's *current* session. Covered by regression tests that fail on `master`. `kill_window` documents the one case it cannot disambiguate (a window *name* containing `:`).

### Tooling — `pyproject.toml`, `src/libtmux/pytest_plugin.py`

- **Add** `pytest-asyncio` (dev + testing); set `asyncio_mode = "strict"` with `asyncio_default_fixture_loop_scope = "function"`, and mark the async test modules explicitly. Inner `pytester` runs keep `-p no:asyncio` (they're isolated and don't read this project's loop-scope config).

### Docs — `docs/experiment/`, `docs/index.md`, `docs/project/public-api.md`

- **Add** an experiment landing page plus a per-module autodoc page; wire `experiment/index` into the toctree; mark `libtmux._experimental.*` not-public. *(Follow-up: a `_resolve` autodoc page is not yet wired in.)*

## Design decisions

- **One substrate, layered API**: everything compiles down to `ir.CommandChain`, so the IR, expressions, async, and connection layers all share one dispatch path and one set of guarantees.
- **Forward refs are one type, two states**: a `PaneRef` is concrete or pending, reusing its existing `.cmd`/`.window`/`.session` namespaces in both states rather than introducing a parallel "deferred" object hierarchy; pending metadata fails closed.
- **The resolver picks the cheapest correct strategy**: a single pane handle resolves in one `{marked}` invocation, several independent handles in the minimum N — chosen automatically from the plan's shape, not by the caller.
- **Sync and async share one core**: the multi-dispatch resolver is a sans-I/O generator with two thin drivers, mirroring how `_async` reuses the sync `to_chain`; the async executor offloads the sync core via `asyncio.to_thread`.
- **Chainability is a real gate**: `to_chain` rejects a non-chainable command with `ChainabilityError`; raw `CommandCall >> CommandCall` composition is the explicit, unchecked escape hatch.
- **Deferred results are two-state**: a chained call has no result of its own until the chain runs; `run_deferred` dispatches once and hands back one resolved handle per command (each reflecting the chain's single merged result, since a `\;` dispatch is not separable per command).
- **Typed, fail-closed targets**: row-bound commands carry `PaneTarget` / `WindowTarget` / `SessionTarget`; `CommandCall` rejects an empty-string target, the snapshot adapter skips rows missing an id, and creation verbs are guarded to their valid scope, so a command cannot silently mis-target.
- **Module names follow common Python conventions**: `ir` for the intermediate representation (mypy `mypyc/ir`, polars `plans/ir`), `plan` for the deferred form (datafusion `LogicalPlan`), `chain` for fold-into-one-dispatch, `_resolve` for the multi-dispatch core, and `_connection` + `*Executor` for the live bridge (django `db/backends`, dagster `Executor.execute(plan)`, stdlib `concurrent.futures.Executor`).

## Test plan

- [x] `uv run ruff check . --fix --show-fixes` — clean
- [x] `uv run ruff format .` — clean
- [x] `uv run mypy` — clean (strict, over `src` + `tests`)
- [x] `uv run pytest --reruns 0` — passes (incl. new doctests, async tests, live-tmux integration covering forward refs / N-dispatch / single-`{marked}`-dispatch, and the targeting regression tests)
- [x] `just build-docs` — builds (experiment toctree + autodoc resolve)

## Scope note

Two commits are separable from the experimental feature, in case you'd prefer them in their own PRs:
- the `Session`/`Pane` targeting fix and its regression tests (a real shipped-behavior fix), and
- a test-only change making the `retry_until` timing tests deterministic under load.

Refs #683. Builds on the API survey in #684.


tony:
### Code review

tony:
### Code review

PR #698, opened 2026-07-03
Fix window-name test for tmux 3.7a/3.7b and add them to CI

tony:
## Summary

- **Fix** the window-name test to assert that `:` and `.` in window names are **accepted**. The old test expected tmux to *reject* them, which only the tmux **3.7** point release ever did — so it fails on tmux **3.7a**/**3.7b**.
- **Add** tmux **3.7a** and **3.7b** to the CI matrix so the grid exercises the 3.7 line's shipped patch releases.
- **Drop** the superseded tmux **3.7** from the matrix — it is the lone release that rejects `:`/`.`, which a pure acceptance assertion can't accommodate.

## Root cause

tmux's window-name sanitization changed around the 3.7 release:

| tmux version | `:` / `.` in window names |
|---|---|
| 3.2a – 3.6 | accepted |
| **3.7** (point release) | **rejected** (`invalid window name`) |
| 3.7a, 3.7b, master | accepted again |

The rejection was introduced just before 3.7 and reverted in 3.7a — upstream tmux commit [`166267c8`](https://github.com/tmux/tmux/commit/166267c8), *"Allow :. in names again, forbidding them is overly pernickety."* So it existed in exactly one release. Verified empirically on tmux 3.7b: `project:frontend` and `app-v1.0` are accepted.

The test gated on `has_gte_version("3.7")`, expecting rejection for all `>= 3.7`. But libtmux's `get_version()` strips letter suffixes (`re.sub(r"[a-z-]", "", version)`), so **3.7, 3.7a, and 3.7b all compare equal to `3.7`**. libtmux cannot distinguish the point release from its patches, so on 3.7a/3.7b the test took the reject-expecting branch and failed with `DID NOT RAISE`.

## Design decisions

- **Assert acceptance, not rejection.** `:`/`.` names work on every tmux libtmux supports except the single superseded 3.7 release. Since `get_version()` can't detect literal 3.7 anyway, an unconditional acceptance assertion is both correct and the most the version API can express.
- **Represent the 3.7 line by 3.7a/3.7b, not 3.7.** A pure acceptance test only stays green if no grid version rejects the names. 3.7a/3.7b are the patch releases users actually run (3.7 was superseded within days) and exercise the same 3.7-family gates (`has_gte_version("3.7")`, `has_version("3.7")`) identically.

## Test plan

- [x] New `test_new_window_name_colon_period_accepted[colon]`/`[period]` pass on local tmux 3.7b
- [x] Full suite green locally (`uv run pytest --reruns 0`) — the only failures were pre-existing capture/timing flakes that pass when re-run serially
- [x] `uv run ruff check .` and `ruff format . --check` clean; `uv run mypy src tests` clean; `just build-docs` succeeds
- [x] CI grid exercises the new tmux 3.7a and 3.7b cells

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.14']
-        tmux-version: ['3.2a', '3.3a', '3.4', '3.5', '3.6', '3.7', 'master']
+        tmux-version: ['3.2a', '3.3a', '3.4', '3.5', '3.6', '3.7a', '3.7b', 'master']
     steps:
       - uses: actions/checkout@v6
 
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,15 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Development
+
+#### tmux 3.7a/3.7b test compatibility (#698)
+
+libtmux's test suite now passes against tmux 3.7a and 3.7b, which CI also
+exercises. A window-name test had assumed the `:`/`.` rejection tmux
+introduced in 3.7, but tmux reverted it in 3.7a as "overly pernickety" — so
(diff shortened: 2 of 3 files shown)


tony:
### Code review

PR #699, opened 2026-07-04
Fix break_pane forcing 'libtmux' window name on tmux 3.7a/3.7b

tony:
## Summary

- **Fix** `Pane.break_pane()` forcing the window name to `libtmux` on tmux **3.7a**/**3.7b** when called without an explicit `window_name` — it now keeps tmux's natural auto-name.
- **Add** `libtmux.common.get_version_str()` — the raw tmux version string with its point-release suffix intact (`"3.7a"`), for telling patch releases apart where `get_version()` can't.
- **Refactor** tmux-version detection into two plain, typed `@functools.cache` helpers; `get_version` output stays byte-identical.

## The bug

`break_pane()` works around a tmux 3.7 NULL-deref (break-pane with no `-n` crashes the server) by injecting a placeholder `-n libtmux` and renaming afterward. The gate was `has_version("3.7")` — but `get_version()` strips the letter suffix for numeric comparison, so it also matched **3.7a/3.7b**, where the crash was already reverted upstream ([tmux `166267c8`](https://github.com/tmux/tmux/commit/166267c8), shipped in 3.7a). With no explicit name, the placeholder leaked as the window name.

Shipped in v0.59.0/v0.60.0. Reproduced on tmux 3.7b: an unnamed `break_pane()` produced a window named `libtmux` instead of the running command's name.

## The fix

Gate the workaround on the *raw* version string being exactly `"3.7"` — which the numeric `get_version()` can't express. `get_version_str()` exposes that raw token, and `break_pane` now compares `get_version_str(...) == "3.7"`, so 3.7a/3.7b (and master, and older tmux) keep tmux's natural window naming.

## Version-cache design

`get_version` and `get_version_str` are two plain `@functools.cache` functions deriving from a shared uncached `_query_version`, each with native `.cache_clear()`/`.cache_info()` — no shared-cache machinery, no `object.__setattr__`, no callable-class wrapper. `get_version` stays byte-identical across every `tmux -V` variation (real releases, `master`, `next-3.8`, OpenBSD, and both error paths); the OpenBSD/no-`-V` synthetic is distinguished by control flow (an internal exception), not a value sentinel that could collide.

## Test plan

- [x] `test_break_pane_no_name_uses_natural_name` — regression test, added as a strict `xfail` then un-xfail'd once the fix landed (the fix commit's CI went red on exactly the 3.7a/3.7b cells, proving it)
- [x] `get_version` verified byte-identical to the prior implementation across all `tmux -V` variations
- [x] `uv run mypy src tests` clean (zero `# type: ignore`); `ruff` clean
- [x] `get_version` / `get_version_str` render in the API docs with signatures
- [x] CI grid green on tmux 3.7a/3.7b (added to the matrix in #698)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,16 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### What's new
+
+#### New helper: `get_version_str()` (#699)
+
+{func}`libtmux.common.get_version_str` returns the running tmux version
+verbatim, keeping the point-release suffix (`"3.7a"`) that
+{func}`~libtmux.common.get_version` strips for numeric comparison. Reach
+for it to distinguish patch releases whose behavior differs -- for
+example telling tmux `3.7` from `3.7a`.
+
 ### Fixes
 
 #### `break_pane()` keeps tmux's default window name on 3.7a/3.7b (#699)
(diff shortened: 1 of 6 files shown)


tony:
### Code review

PR #705, opened 2026-07-11
Complete the from_env family: know which pane, window, session and server you're running in

### LINKED ISSUE #704: No self-location API (TMUX_PANE), and the socket_path derivation in Server.__init__ is unreachable dead code
*Filed against **tmux-python/libtmux** (v0.61.0). Found while building an MCP server on libtmux; both items force every downstream consumer to re-implement the same two things.*

## 1. `Server.__init__`'s `socket_path` derivation is unreachable

The `TMUX_TMPDIR`-based `socket_path` auto-derivation block in `Server.__init__` can never execute — its conditions are mutually contradictory. A few lines earlier the local is rebound (`socket_name = self.socket_name or "default"`), so `self.socket_name is None` implies the local `socket_name == "default"`, and the two guards can never both hold.

- [`src/libtmux/server.py#L198`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L198)

**Consequence:** `server.socket_path` is **always `None`** unless the caller passed `socket_path=` explicitly. Verified: `Server(socket_name="x").socket_path is None`, and bare `Server().socket_path is None`.

This is why downstream code has to re-resolve the socket path by hand (re-deriving `TMUX_TMPDIR` + euid + socket name, then falling back to shelling out `tmux display-message -p '#{socket_path}'`). Every consumer that needs to *prove* which server it is talking to writes this twice.

**Suggested fix:** either repair the condition so the derivation runs, or delete the dead block and document that `socket_path` is populated only when supplied. Repairing it is better — a `Server` that knows its own socket path lets consumers do socket-identity checks without a subprocess.

## 2. There is no self-location API

`TMUX_PANE` does not appear **anywhere** in the package (verified by ripgrep over the whole tree at v0.61.0). There is no `Server.from_env()`, no `current_pane()`, no equivalent. The only place libtmux touches `$TMUX` at all is `Server.new_session`, which **deletes** it before shelling out (so a nested `new-session` doesn't inherit the parent server) and restores it in a `finally` — it never *parses* it.

- [`src/libtmux/server.py#L2278`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L2278)

So any process running **inside** a tmux pane — an agent, an MCP server, a CLI tool — must hand-roll:

```python
pane_id = os.environ.get("TMUX_PANE")            # "%65"
socket_path, server_pid, session_id = os.environ["TMUX"].split(",")
```

…and then find a way to turn that into a `Pane`. The obvious route is a trap:

| approach | cost | why |
|---|---|---|
| `Pane.from_pane_id(...)` | **28.4 ms** | delegates to `neo.fetch_obj` → full `list-panes -a` with the wide `-F` template, then a **Python-side** linear scan ([`neo.py#L787`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/neo.py#L787)) |
| `server.panes.filter(pane_id=...)` | **28.7 ms** | `QueryList.filter/get` are pure-Python predicates over an already-enumerated list; they never push down to tmux ([`query_list.py#L513`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/_internal/query_list.py#L513)) |
| `Server.search_panes(filter=...)` | 9.8 ms | does push tmux's native `-f`, but still runs `list-panes -a` with the wide template ([`server.py#L2563`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L2563)) |
| **`Pane(server=srv, pane_id=...)` + `display_message`** | **3.3 ms** | dataclass construction is free, then **one** targeted `display-message` |

The fast path works only because of two facts a caller has to *discover*: `Pane` is a dataclass whose only required field is `server` ([`neo.py#L389`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/neo.py#L389)), and `Pane.cmd` auto-injects `-t <pane_id>` ([`pane.py#L214`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/pane.py#L214)). Nothing signposts this, so the natural-looking API (`from_pane_id`) is ~9× slower than the right one.

Note also that `Client.attached_pane` / `Session.active_window` / `Window.active_pane` are **not** substitutes: they resolve *"currently focused"* via `list-clients` / `pane_active=1`, not *"the pane this process is running in"*. A headless process has no client row at all.
- [`src/libtmux/client.py#L170`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/client.py#L170)
- [`src/libtmux/session.py#L525`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/session.py#L525)

### Proposal

```python
libtmux.Server.from_env()   # parse $TMUX -> Server(socket_path=...)   (no subprocess)
libtmux.current_pane()      # read $TMUX_PANE -> Pane(server=..., pane_id=...)  (no subprocess)
```

Both are a handful of lines, need no tmux round-trip, and would let every downstream consumer stop re-implementing the same env parse. Bonus: `socket_path` and `pid` are already **universal** format tokens declared on `Obj` ([`neo.py#L146`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/neo.py#L146)), so a single `display-message` can return the server's socket path and PID — exactly the first two components of `TMUX=<socketpath>,<pid>,<session>` — giving a zero-extra-cost proof that a pane really belongs to the caller's server.

## 3. Two smaller papercuts found alongside

**(a) `Server.start_server()` leaves a stale socket file.** It runs `tmux start-server`, but with `exit-empty` on (the default) the sessionless daemon exits immediately — `is_alive()` still returns `False`, yet **a stale socket file remains on disk** and survives `kill()`. So socket-file existence is not a valid liveness test, and `start_server()` alone cannot "ensure a server".
- [`src/libtmux/server.py#L1448`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L1448)

Suggested: document this, or have `start_server()` verify liveness and raise/warn.

**(b) Error-handling asymmetry between `.sessions` and `.panes`.** `Server.sessions` and `Server.clients` swallow **every** `LibTmuxException` and return an empty `QueryList` — so a *permission error* is indistinguishable from *"no sessions"*. `Server.panes`/`Server.windows` instead go through `_fetch_or_empty`, which swallows only daemon-not-up errors (classified by `_is_daemon_not_up_error`) and **re-raises** everything else. Confirmed: a "File name too long" socket error propagates out of `.panes` but not out of `.sessions`.
- [`src/libtmux/server.py#L2372`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L2372) (swallow-all)
- [`src/libtmux/server.py#L79`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L79) (`_fetch_or_empty`)

Suggested: route `.sessions`/`.clients` through `_fetch_or_empty` too, so emptiness means the same thing everywhere.

**(c) `_is_daemon_not_up_error` is private but universally needed.** It is the canonical classifier for both tmux phrasings (`"no server running"` and `"error connecting to … (No such file or directory)"`). Downstream consumers currently string-match independently. Consider exporting it, or a public `ServerStatus` enum alongside `is_alive()`.
- [`src/libtmux/server.py#L58`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L58)

**(d) `EnvironmentMixin.set_environment` raises a bare `ValueError`** on tmux stderr, bypassing the `LibTmuxException` hierarchy. Downstream error mappers key on `LibTmuxException`, so a routine "no tmux server running" escapes as an *unexpected* error.
- [`src/libtmux/common.py#L115`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/common.py#L115)

## Environment

- libtmux: 0.61.0
- tmux: 3.7b
- Python: 3.14.0

---

**Cross-references** — filed together from a single audit of `v0.1.0a17`. The self-location gap is the root cause; the rest are what it exposed.

- tmux-python/libtmux-mcp#92 — **[bug]** documented self-location recipe is unexecutable (`filters={"is_caller": true}` → `[]`)
- tmux-python/libtmux-mcp#93 — **[feat]** `where_am_i` self-location primitive + self-first relational routing *(root fix)*
- tmux-python/libtmux-mcp#94 — **[bug]** caller's socket advertised but never targeted; `TMUX` env race vs. the threadpool
- tmux-python/libtmux-mcp#95 — **[perf]** `list_*` unbounded, unscoped, unprojected, excluded from the response limiter
- tmux-python/libtmux-mcp#96 — **[feat]** invocation context (in tmux? server running?) + explicit tmux-server auto-start
- tmux-python/libtmux#704 — **upstream:** no self-location API (`TMUX_PANE`); unreachable `socket_path` derivation

*(← this issue: tmux-python/libtmux#704)*

tony:
Closes #704.

## Why

A process running inside tmux had no way to ask **where it is**. tmux hands every pane it spawns `TMUX` (`socket_path,server_pid,session_id`) and `TMUX_PANE`, but libtmux read neither — `TMUX_PANE` appeared nowhere in the package. Callers parsed the environment by hand and then reached for `Pane.from_pane_id()`, which lists every pane on the server and filters in Python.

## What

`Server.from_env()` · `Session.from_env()` · `Window.from_env()` · `Pane.from_env()` — each returns the object the current process is running inside. All accept an optional `env` mapping (defaulting to `os.environ`), which keeps them injectable and lets the doctests run deterministically inside *and* outside tmux.

```python
>>> Server.from_env({"TMUX": "/tmp/tmux-1000/default,123,$0"})
Server(socket_path=/tmp/tmux-1000/default)
```

`exc.NotInsideTmux` — all four raise rather than guess. Silently answering with someone else's pane is worse than failing.

**They report what _contains_ you, not what is focused.** Implementing `Window.from_env()` as `session.active_window` would be wrong for any background pane; there's a test with a deliberately non-active window pinning that.

**No `Client.from_env()`.** A tmux client is an attached terminal identified by its tty; a pane process owns the *pane's pty*, not the client's tty, and tmux exports no client id. The cardinality is 0..N with **0 as the default** (detached session, CI, a `send-keys` script). Even `ControlMode` — the one process that genuinely *is* a client — recovers its identity by PID-correlating `list-clients`, not from the environment. If the underlying need is real, the honest API is `Session.clients -> QueryList[Client]`.

## The server answers; `$TMUX` only breaks ties

`$TMUX` carries a session id and it is tempting to just read it. It is **wrong**: tmux freezes the variable into the pane's environ at spawn (`environ.c` / `spawn.c`) and never revises it, so after a `move-window` it still names the old session.

But resolving purely from the pane is *also* wrong — subtly. `link-window` puts one window in several sessions, and tmux then lists the pane **once per session, ordered by session name**. Resolving from the pane alone therefore answers with whichever session sorts last:

| origin / dest session names | resolved session |
|---|---|
| `aaa_origin` / `zzz_dest` | **dest** ✗ |
| `zzz_origin` / `aaa_dest` | origin ✓ *(by luck)* |

Same topology, different names, different answer. So: **the live server is ground truth; `$TMUX` breaks the tie** when several sessions genuinely hold the pane, because it alone records which one the process was *spawned* under. When it names a session that no longer holds the pane, the window moved and the server wins. An unusable id (`-1`, a killed session, anything unparseable) is simply ignored. Still 2 tmux calls.

## Also: a guard on `Pane.session`

`Pane.session` resolves through `Pane.window` — two round-trips — when the pane already carries a `session_id` that *looks* like it would do. That shortcut is a **correctness bug**: a pane's cached `session_id` goes stale when its window moves; its `window_id` does not. Verified — swapping the hop for the cached id turns the new test red. Documented and pinned so the next reader doesn't "optimize" it away.

## Fix: unreachable `socket_path` derivation

The `TMUX_TMPDIR` derivation in `Server.__init__` could never run: the local `socket_name` is rebound to `self.socket_name or "default"`, so `self.socket_name is None` implies it *equals* `"default"` — which the trailing `socket_name != "default"` guard rejects. Dead since `aa7a3ae2` ("fix launching default server"), and **reviving it would be wrong**: `Server.cmd` and `neo.fetch_objs` each emit `-L` *and* `-S`, so a `Server` with a `socket_name` plus a derived `socket_path` would target the server twice. Removed, with a test pinning the contract.

## Deliberately not in scope

#704 also floated routing `Server.sessions` through `_fetch_or_empty`. `AGENTS.md` documents the current behavior as an intentional contract ("list-shaped accessors are lenient by default… Empty-on-tmux-error stays the default; raise is opt-in"), so that suggestion is **withdrawn**.

## Testing

Written test-first. Covers: `$TMUX` parsing (including the bare id tmux actually writes — `47`, not `$47`), `os.environ` fallback, live-server round-trip, **move-window**, **link-window**, non-active windows, unusable spawn ids, and both not-inside-tmux paths. Doctests execute against live fixtures — no `+SKIP`.

`ruff format` · `ruff check` · `mypy src tests` · full `pytest` (**1371 passed**) · `just build-docs` — all clean.

> **Pre-existing, unrelated:** `docs/topics/automation_patterns.md[4]` is a timing-sensitive doctest that fails intermittently on clean `master` too (verified: 2 of 3 runs at `--reruns 0`, and with the default `--reruns 2`). Not introduced here — likely worth its own issue.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,31 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### What's new
+
+#### Find where you are running (#705)
+
+A program started inside tmux can now ask where it is.
+{meth}`Server.from_env() <libtmux.Server.from_env>`, {meth}`Session.from_env()
+<libtmux.Session.from_env>`, {meth}`Window.from_env() <libtmux.Window.from_env>`
+and {meth}`Pane.from_env() <libtmux.Pane.from_env>` each return the object the
+current process is running inside, so a script, a hook, or an agent can pick up
+the hierarchy from wherever it happens to be. Outside tmux they raise
+{exc}`~libtmux.exc.NotInsideTmux` rather than guess, and each accepts an
+environment mapping in place of the process environment, which keeps them
+straightforward to test. See {ref}`self-location`.
+
+They report what *contains* the caller, not what is focused — a pane running in
+the background gets its own window back, not the one someone happens to be
+looking at. And the server gives the answer, not the environment: the session id
+tmux writes into `TMUX` goes stale the moment a window is moved, so resolution
+anchors on the pane, which tmux keeps answering for live.
+
+There is no `Client.from_env()`. A tmux client is an attached terminal, and a
+pane is not owned by one — none may be attached, or several may be, each with
+its own view. tmux exports no client id into a pane, so there is nothing to read
+back.
+
 ## libtmux 0.61.0 (2026-07-04)
 
 libtmux 0.61.0 hardens support for the tmux 3.7 patch line. It fixes
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ libtmux is a typed Python API over [tmux], the terminal multiplexer. Stop shelli
 
 - Typed, object-oriented control of tmux state
 - Query and [traverse](https://libtmux.git-pull.com/topics/traversal/) live sessions, windows, and panes
+- [Locate yourself](https://libtmux.git-pull.com/topics/self_location/) from inside a pane with `from_env()`
 - Raw escape hatch via `.cmd(...)` on any object
 - Works with multiple tmux sockets and servers
 - [Context managers](https://libtmux.git-pull.com/topics/context_managers/) for automatic cleanup
@@ -225,6 +226,27 @@ Window(@... ...:..., Session($... ...))
 >>> pane.window.session
 Session($... ...)
 ```
+
+### Know where you're running
+
+[**Learn more about Locating yourself**](https://libtmux.git-pull.com/topics/self_location/)
+
+Code *running inside* a pane — a script in a split, a tmux hook, an agent — can ask where it is. tmux writes `TMUX` and `TMUX_PANE` into every pane it spawns, and `Server`, `Session`, `Window`, and `Pane` each read them back:
+
+```python
+>>> socket_path = server.cmd(
+...     "display-message", "-p", "-t", session.session_id, "#{socket_path}"
+... ).stdout[0]
+>>> monkeypatch.setenv("TMUX", f"{socket_path},1,{session.session_id}")
+>>> monkeypatch.setenv("TMUX_PANE", pane.pane_id)
+
+>>> Pane.from_env()
+Pane(%... ...)
+>>> Session.from_env().session_name == session.session_name
+True
+```
+
+In a real pane tmux has already set those two variables, so `from_env()` takes no arguments and there is nothing to arrange — this README is not running in a pane, so the example sets them first. Outside tmux there is no pane to return, and `from_env()` raises `NotInsideTmux`.
 
 ## Core concepts
 
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -63,17 +63,30 @@ requires them to actually execute — `testpaths` includes `docs/`, so
 pytest runs every one. Lead with a small, runnable example early rather
 than after paragraphs of prose; libtmux is code-first.
 
-- Use the `doctest_namespace` fixtures — `server`, `session`, `window`,
-  `pane` (and `Server` / `Session` / `Window` / `Pane` / `Client`) —
-  instead of building a server by hand.
+- Use the `doctest_namespace` fixtures instead of building a server by
+  hand. `conftest.py` seeds the namespace with `server`, `session`,
+  `window`, `pane`, the `Server` / `Session` / `Window` / `Pane` /
+  `Client` classes, `ControlMode` and `control_mode`, `monkeypatch`,
+  and `request`.
 - Fence a `>>>` session as a ```` ```python ```` block, and reach for
   `# doctest: +ELLIPSIS` when output varies (ids like `@1`, `$2`,
   socket names). Use a ```` ```console ```` block for shell commands at
   a `$` prompt.
-- The code blocks on a page share one doctest session, so a later
-  block can use a `pane` an earlier block created. That makes their
-  **order load-bearing**: never reorder, add, or drop a code block when
-  you reshape the prose around it.
+- **Every code block on a page is an independent doctest.** Blocks do
+  not share a session: each one gets a fresh copy of the namespace *and*
+  a fresh tmux server. A later block cannot use a `pane` an earlier
+  block created — the name is simply not defined there. So **write every
+  block self-contained**. The fixtures above are re-seeded for each
+  block, which makes that cheap: no block needs an import or a setup
+  preamble to get a `server`, a `session`, or a `pane`. Order is
+  load-bearing for the reader's narrative, not for state — reorder, add,
+  or drop a block as the prose demands, and keep the story coherent.
+- Two of those names are not quite what they look like. `Server` is a
+  `TestServer` partial, not the real class; write
+  `>>> from libtmux.server import Server` when an example needs the
+  class itself. And the fixture server is created with a socket *name*,
+  so `server.socket_path` is `None` — an example that needs the path
+  must ask tmux for it with `display-message -p '#{socket_path}'`.
 
 ## What stays precise
 
@@ -109,8 +122,8 @@ another link useful.
 Do not rely on a later reference section to satisfy the first-mention rule. If
 the first occurrence would be a heading, grid-card teaser, or introductory
 sentence, link that occurrence or retitle the heading so the first prose mention
-can carry the link. Leave command examples, code blocks, Mermaid node labels,
-and literal configuration values as code; link the surrounding prose instead.
+can carry the link. Leave command examples, code blocks, and literal
+configuration values as code; link the surrounding prose instead.
 
 ## A page that does this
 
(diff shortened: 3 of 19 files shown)


tony:
Superseded by #713 and #714.

A clean-room rebuild found two defects in this branch that were worth starting over for.

**It contradicted itself.** For a window linked into two sessions, `Session.from_env()` used a tie-break based on the session id in `$TMUX`, while `Window.from_env().session` and `Pane.from_env().session` traversed and got a different, arbitrary session. Worse, the tie-break also disagreed with tmux: for the same pane, `tmux display-message -t` reported one session and `Session.from_env()` insisted on another.

**One constructor escaped the exception hierarchy.** On a dead server, `Session.from_env()` raised `subprocess.CalledProcessError` (via `Server.raise_if_dead()`), while every sibling raised a `LibTmuxException`.

Both had the same root cause, and it was not in `from_env` at all: the four point lookups scanned the whole server with `-a` and kept the last matching row, which tmux orders by session *name*. No amount of cleverness inside `from_env` could fix that, because `Pane.session` traverses through `Window.from_window_id` to get its answer.

#713 fixes the resolution itself — the lookups now name the object with `-t` and let tmux's `cmd_find` resolve it, so libtmux agrees with tmux (and #710 is fixed for everyone, not just `from_env` callers). #714 then builds `from_env` on top, and it no longer needs to parse the session id out of `$TMUX`, invent a tie-break, or reach for `raise_if_dead()`. The `$TMUX` session field is never read.

The docs, the exception, and the `Server.__init__` dead-code removal from this branch all carry forward. Reading the *spawn* session as explicit provenance — a genuinely separate question from "where am I now?" — is proposed in #712.

PR #709, opened 2026-07-11
Prevent shell-echo races in automation docs

### LINKED ISSUE #706: docs(automation_patterns): the wait helpers never wait — the marker matches the command echo
## Bug

Both wait helpers taught in `docs/topics/automation_patterns.md` return **before the command they are waiting on has produced any output**, because the shell echoes the command line and the marker is matched *in that echo*.

Same root cause as #654, but in the **documentation** — so it is the pattern readers copy into their own code.

The two helpers fail differently, and the *quieter* one is the worse one:

| helper | failure |
|---|---|
| [`wait_for_output`](https://github.com/tmux-python/libtmux/blob/v0.61.0/docs/topics/automation_patterns.md?plain=1#L127) | **always returns `True` immediately.** `timeout` and `poll_interval` are dead parameters. Its doctest passes *vacuously* — it would pass if the command never ran at all. |
| [`capture_after_marker`](https://github.com/tmux-python/libtmux/blob/v0.61.0/docs/topics/automation_patterns.md?plain=1#L190) | its poll loop exits on the first capture, so `timeout=2.0` never waits; the only real wait is a bare `time.sleep(0.3)`. Visibly flaky — `automation_patterns.md[4]` fails intermittently. |

## Root cause

`send-keys` types the command into the pane, and the shell echoes it back before running it. The echoed line **contains the marker string**, so a substring match over `capture_pane()` succeeds instantly.

[`docs/topics/automation_patterns.md#L137-L139`](https://github.com/tmux-python/libtmux/blob/v0.61.0/docs/topics/automation_patterns.md?plain=1#L137-L139):

```python
>>> monitor_pane.send_keys('sleep 0.2; echo "READY"')
>>> wait_for_output(monitor_pane, 'READY', timeout=2.0)
True
```

`READY` is in `sleep 0.2; echo "READY"` — the command itself. The helper matches the echo, not the output.

## Reproduction

The helper verbatim from the page, with the sleep raised so the race is unmissable:

```python
pane.send_keys('sleep 2.0; echo "READY"')
got = wait_for_output(pane, "READY", timeout=5.0)   # helper copied from the docs
```

Observed:

```
wait_for_output(...) returned True after 0.02s
the command needs >= 2.0s to print READY
screen at that moment:
   'sleep 2.0; echo "READY"'      <- only the echo; no output yet
```

**Observed:** `True` in 0.02 s.
**Expected:** blocks ~2 s, then `True` — or `False` on timeout.

`capture_after_marker` fails the same way: its first captured line containing `MARKER` is index 0, the echoed command.

## Suggested fix

**Teach `tmux wait-for` rather than polling for a string.** libtmux already wraps it — [`Server.wait_for()`](https://github.com/tmux-python/libtmux/blob/v0.61.0/src/libtmux/server.py#L571) — and it is race-free *by construction*: no marker, no poll interval, no sleep.

```python
pane.send_keys('sleep 2.0; echo "captured data"; tmux wait-for -S done')
server.wait_for('done')            # blocks until the command actually finished
result = pane.capture_pane()
```

There is **no lost-wakeup race** even if the command finishes before the wait is issued. From tmux's `cmd-wait-for.c`:

```c
/* signal: nobody waiting yet -> remember it */
if (TAILQ_EMPTY(&wc->waiters) && !wc->woken) { wc->woken = 1; return (CMD_RETURN_NORMAL); }

/* wait: already woken -> return at once */
if (wc->woken) { cmd_wait_for_remove(wc); return (CMD_RETURN_NORMAL); }
```

Inside a pane `$TMUX` is already set, so a bare `tmux wait-for -S done` reaches the right server with no extra wiring.

If a string-matching example is still wanted — reasonable for output you do *not* control — it must not be matchable against the echo. Either match per line and skip the echoed command (the fix proposed in #654), or use a sentinel that cannot appear in the command line, e.g. assemble it at runtime so the literal is never typed.

## Tests

The page's doctests currently prove nothing, so they need to become real assertions:

1. `wait_for_output` against a command that sleeps **longer than the poll interval** must take at least that long to return — that fails today.
2. `wait_for_output` must return `False` on timeout when the text never appears (today it cannot, since it matches the echo).
3. `capture_after_marker` must return the command's *output*, not the tail of the echoed line.

## Related

- #654 — same root cause, in a test rather than the docs
- #484, #456 — adjacent `capture_pane` reliability

## Environment

- libtmux: v0.61.0 (both helpers unchanged since the tag)
- tmux: 3.7b
- Python: 3.14

tony:
## Summary

Replace echo-prone completion polling with tmux `wait-for` channels when automation controls the command and does not need a timeout.

For bounded waits, retries, queues, and state machines, match echo-safe complete output lines against monotonic deadlines. Stop retry and queue sequencing when a timed-out foreground process may still own the pane.

Update every affected executable example and the page guidance so doctests prove command output rather than shell input echo.

## Verification

- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run mypy .`
- `just build-docs`
- `uv run py.test --reruns 0 -q docs/topics/automation_patterns.md` (`16 passed`)
- `uv run py.test --reruns 0 -q` (`1342 passed, 1 skipped`)

Closes #706

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,19 @@ _Notes on the upcoming release will go here._
 
 ### Documentation
 
+#### The automation recipes no longer race the shell's echo (#709)
+
+Every wait helper in {ref}`automation-patterns` returned the moment the command was
+*sent*, not when it finished. {meth}`~libtmux.Pane.send_keys` types into the pane, so
+the shell echoes it straight back, and the helpers matched their completion marker
+against that echo. A copied `wait_for_output(pane, 'READY')` returned `True` in
+milliseconds, before the command had printed anything, and its timeout branch was
+unreachable. `check_for_errors` had the same fault in reverse: it reported a crash for
+any healthy command whose own arguments happened to mention an error.
+
+The recipes now synchronize through {meth}`~libtmux.Server.wait_for` channels where the
+command is yours to change, and where a poll is genuinely needed they match complete
+output lines that cannot also occur in the echoed input.
 #### Cleaner `from_env` examples (#719)
 
 The rendered examples for {meth}`Pane.from_env() <libtmux.Pane.from_env>` and
--- a/docs/topics/automation_patterns.md
+++ b/docs/topics/automation_patterns.md
@@ -149,34 +149,66 @@ True
 
 ### Detecting errors in output
 
-Waiting for success is only half the job — you also want to notice failure. The same
-capture-and-scan approach works for spotting error patterns, so you can bail out
-early instead of timing out on a command that already crashed.
+Waiting for success is only half the job — you also want to notice failure, so you
+can bail out early instead of timing out on a command that already crashed.
+
+Scan the command's **output**, never the whole pane. The shell echoes what you sent
+into the same buffer, so a pane-wide substring search reports a crash for
+`mytool --log-level=ERROR` before the tool has printed anything at all — it is
+matching your own command line. That is the same trap the polls above avoid, in the
+one place it is easiest to walk into.
+
+So have the command tag the lines it owns, and scan only those. The tag has to be one
+that cannot begin a line of the echoed command — and a long command *wraps*, so "it
+is not at the start of my command" is not good enough. Let the shell assemble it:
+below, `printf` builds `out:` from a format and a separate argument, so the literal
+never appears in what you typed, wrapped or not.
 
 ```python
 >>> error_window = session.new_window(window_name='error-check', attach=False)
 >>> error_pane = error_window.active_pane
 
->>> def check_for_errors(pane, patterns=None):
-...     """Check pane output for error patterns."""
+>>> def command_output(pane, tag='out:'):
+...     """The lines the command tagged as its own."""
+...     return [
+...         line.removeprefix(tag)
+...         for line in pane.capture_pane()
+...         if line.startswith(tag)
+...     ]
+
+>>> def check_for_errors(lines, patterns=None):
+...     """Return the first error pattern found in *lines*, or None."""
 ...     if patterns is None:
 ...         patterns = ['Error:', 'error:', 'ERROR', 'FAILED', 'Exception']
-...     output = '\\n'.join(pane.capture_pane())
-...     for pattern in patterns:
-...         if pattern in output:
-...             return pattern
+...     for line in lines:
+...         for pattern in patterns:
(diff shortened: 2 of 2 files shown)


tony:
Pushed a follow-up: `check_for_errors` was still carrying the bug this PR exists to remove.

The rest of the page now waits on `wait-for` channels and matches whole output lines, but that one helper was left doing `pattern in '\n'.join(pane.capture_pane())`. `send_keys` types the command, the shell echoes it back into the same buffer, so it reported a crash for any healthy command whose own arguments mention an error. Reproduced against live tmux before and after:

```
pane contents: ["true --log-level=ERROR; printf 'out%s%s\n' : ok; tmux wait-for -S ch", 'out:ok']

  before (scans whole pane):     'ERROR'   <- healthy command reported as a crash
  after  (scans tagged output):  None
```

The helper now takes the command's own output lines rather than a pane, so it cannot be handed the echo in the first place.

One detail worth flagging, because it caught me: **the tag has to survive wrapping.** My first attempt sent `printf 'out:%s\n' ok` and filtered on `line.startswith('out:')`. That passed — but only by luck. A long command wraps, and the wrapped continuation of the echo is its own line, so a tag that merely sits mid-line in what you typed can still end up starting one. The fix is to let `printf` assemble the tag from a format and a separate argument (`printf 'out%s%s\n' : ok`), which keeps the literal out of the echo entirely, wrapped or not. That is the same trick the page already uses for `printf 'Success%s\n' '!'`, just applied to the prefix instead of the payload.

Also added a `CHANGES` entry under `### Documentation` — the racy recipes shipped in 0.61.0, so they pass the Published-Release Test.

CI green on all 8 tmux versions.

**Scope note:** `Closes #706` is now earned *for the page #706 names*, but the identical pattern is still live on two other pages — `docs/topics/pane_interaction.md:500` and `:519-521`, and `docs/topics/workspace_setup.md:399`. Filed separately so it isn't lost when this merges.

tony:
Superseded by #734, which carries this work forward.

Both PRs modify the same two files, and the resulting `docs/topics/automation_patterns.md` is identical apart from four lines #734 adds. The channel-signalling rewrite and the whole-line comparisons that replace joined-substring searches are the same in both.

Closing in favour of #734 to avoid a conflict between them; #735 is stacked on #734.

Issue #726, opened 2026-07-29
`retry_until`'s zero-arg predicate forces a loop-variable closure that ruff `B023` and mypy disagree about


tony:
*Filed against **tmux-python/libtmux** v0.62.0. Found while writing tested documentation examples that wait on several panes. Related to but distinct from #376, which is about `retry_until`'s failure message rather than its signature.*

## What happens

`retry_until` takes a zero-argument predicate, so waiting on anything from a loop forces a closure over the loop variable. That closure is then rejected by this repo's own lint configuration, and the standard fix for it is rejected by this repo's own type checker. Every such call site ends up carrying a `# type: ignore[misc]`.

- [`src/libtmux/test/retry.py#L22-L28`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/test/retry.py#L22-L28) — the signature

```python
def retry_until(
    fun: Callable[[], bool],
    seconds: float = RETRY_TIMEOUT_SECONDS,
    *,
    interval: float = RETRY_INTERVAL_SECONDS,
    raises: bool | None = True,
) -> bool:
```

## Recreation

Write the natural thing:

```python
from __future__ import annotations

from libtmux.pane import Pane
from libtmux.test.retry import retry_until


def wait_for_each(panes: list[Pane]) -> None:
    for pane in panes:
        retry_until(lambda: any(pane.capture_pane()))
```

`uv run ruff check` — `B023`, which is enabled repo-wide via `extend-select = ["B", ...]` in `pyproject.toml`:

```
B023 Function definition does not bind loop variable `pane`
    retry_until(lambda: any(pane.capture_pane()))
                             ^^^^
```

Apply the documented fix for `B023`, a default-argument binding:

```python
        retry_until(lambda pane=pane: any(pane.capture_pane()))
```

`uv run mypy --strict`:

```
error: Cannot infer type of lambda  [misc]
```

So the lint-clean form does not type-check and the type-clean form does not lint. The only way out at the call site is `# type: ignore[misc]`, which is what the two call sites in `tests/docs/howto/send_keys_to_every_pane.py` on `master` currently carry.

Note that in this particular shape `B023` is a false positive — `retry_until` calls the predicate synchronously before the next iteration, so late binding is harmless. But `B023` cannot know that from the signature, and neither can a reader.

## What a fix needs

The cause is that a zero-argument predicate has nowhere to put the thing being waited on. Options, in descending order of preference:

[...]

Issue #725, opened 2026-07-29
`Server` does not validate socket path length; overrun surfaces as a raw tmux `File name too long`


tony:
*Filed against **tmux-python/libtmux** v0.62.0. Found while writing tested documentation examples that place sockets under a temporary directory.*

## What happens

`Server(socket_path=...)` accepts any path. UNIX domain socket paths are capped by `sun_path` in `struct sockaddr_un` — 108 bytes on Linux, 104 on macOS — so a path longer than that cannot work, but nothing says so until the first tmux command runs and fails with a raw tmux error.

- [`src/libtmux/server.py#L185-L191`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/server.py#L185-L191) — `socket_path` is stored with no validation

```python
        if socket_path is not None:
            self.socket_path = socket_path
```

The path that fails is usually not one the caller typed. It is `tmp_path` from a pytest fixture, an XDG runtime dir, a nested worktree, or a CI checkout under a long workspace prefix — so the number the user needs (how many bytes over) is exactly the number the error does not give them.

## Recreation

```python
import os, pathlib, tempfile
root = pathlib.Path(tempfile.mkdtemp(prefix="repro-"))
deep = root.joinpath(*["d" * 40] * 4) / "sock"
deep.parent.mkdir(parents=True, exist_ok=True)
print("path length:", len(str(deep)), "(UNIX socket limit is ~107)")
from libtmux.server import Server

server = Server(socket_path=str(deep))
print("constructed fine; socket_path =", server.socket_path)
try:
    server.new_session(session_name="demo")
except Exception as exc:
    print(f"{type(exc).__module__}.{type(exc).__name__}: {exc}")
```

Observed on libtmux v0.62.0 / tmux 3.7b:

```
path length: 188 (UNIX socket limit is ~107)
constructed fine; socket_path = /tmp/repro-.../dddd.../sock
libtmux.exc.LibTmuxException: new-session: error connecting to /tmp/repro-.../sock (File name too long)
```

Two problems in that outcome. The construction that was already doomed succeeded, and the diagnosis arrives as a passed-through `File name too long` from tmux at some arbitrary later call — attributed to `new-session`, which is not what is wrong.

## Prior art in this repo

[...]

Issue #724, opened 2026-07-29
`Server.__exit__` kills any live server, including one it did not start


tony:
*Filed against **tmux-python/libtmux** v0.62.0. Found while writing tested documentation examples that isolate tmux under a temporary `TMUX_TMPDIR`.*

## What happens

`with Server(...) as server:` kills the server on exit whenever it is alive, unconditionally. There is no `kill_on_exit=False`, no "only kill what I started", and no way to use the context manager purely for scoping.

- [`src/libtmux/server.py#L267-L285`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/server.py#L267-L285) — `__exit__`

```python
        if self.is_alive():
            self.kill()
```

A `Server` is a *handle*, not a connection, and the same handle addresses a server whether or not this process started it. So the context manager cannot tell "the server I just booted" from "the server the user has been working in all day", and it destroys both.

The sharpest form: `with Server() as server:` — no arguments, so the default socket — **kills the reader's running tmux and every session in it**. That is a plausible thing to write after reading the class docstring, and the failure is silent and total.

## Recreation

```python
import os, tempfile
os.environ["TMUX_TMPDIR"] = tempfile.mkdtemp(prefix="repro-")
from libtmux.server import Server

Server(socket_name="preexisting").new_session(session_name="important-work")
print("before:", [s.session_name for s in Server(socket_name="preexisting").sessions])

with Server(socket_name="preexisting") as server:
    print("inside:", server.is_alive())

print("after :", Server(socket_name="preexisting").is_alive(), "<- the session is gone")
```

Observed on libtmux v0.62.0 / tmux 3.7b:

```
before: ['important-work']
inside: True
after : False <- the session is gone
```

The `with` block did nothing except enter and exit, and a session it never created is gone. Drop the `TMUX_TMPDIR` line and the same script destroys the real default server.

## Note on the sibling context managers

`Session`, `Window` and `Pane` have the same shape — `__exit__` kills the object if it still exists — but the blast radius there is bounded by what the object *is*, and killing a session you were handed is at least proportionate. A server is different in kind: it is shared, process-wide, and typically not yours.

[...]

PR #727, opened 2026-07-29
Resolve the socket path in Server's repr instead of hard-coding /tmp

tony:
## Summary

- **Fix** `repr(Server())` naming a socket the object is not talking to. The fall-through branch hard-coded `/tmp/tmux-<euid>/default`, but tmux resolves its socket directory from `$TMUX_TMPDIR`, so under a test harness, a sandbox, a container, or anyone with `TMUX_TMPDIR` exported the repr was simply wrong.
- **Add** `libtmux._internal.env.resolve_socket_path()`, which computes `tmux-<euid>/<socket_name or "default">` under `$TMUX_TMPDIR`, falling back to `/tmp`.
- **Replace** the inline copy of that computation in `pytest_plugin._reap_test_server`, which carried a comment explaining it had to recompute what `Server` would not tell it. Both callers now share one resolver.
- **Remove** a `getattr(self, 'socket_name', 'default')` in the `socket_name` branch that ran after `self.socket_name is not None` had already been established.

Display-only: which socket the object talks to is unchanged.

## Why the repr matters

A repr is what appears in tracebacks, `--showlocals` banners, logs, and `print(server)` — precisely where someone is trying to work out which server they are looking at. Naming the wrong socket there is worse than naming none.

## Before / After

Under a `TMUX_TMPDIR`:

```python
repr(server)       # Server(socket_path=/tmp/tmux-1000/default)
server.cmd("display-message", "-p", "#{socket_path}").stdout[0]
                   # /tmp/repro-JxuS/tmux-1000/default
```

```python
repr(server)       # Server(socket_path=/tmp/repro-JxuS/tmux-1000/default)
server.cmd("display-message", "-p", "#{socket_path}").stdout[0]
                   # /tmp/repro-JxuS/tmux-1000/default
```

With `TMUX_TMPDIR` unset the output is byte-identical to before, which is why the three shipped doctests that assert this repr needed no edit.

## Design decisions

**`$TMUX_TMPDIR` or `/tmp`, and nothing else.** `TMPDIR` is deliberately not consulted, because tmux does not consult it either — `man tmux`: *"the sockets are all created in a directory tmux-UID under the directory given by TMUX_TMPDIR or in /tmp"*. Confirmed empirically on tmux 3.7b: `TMPDIR=<dir> tmux -L probe` still lands in `/tmp/tmux-1000/probe`. Consulting `TMPDIR` would have broken the existing doctests on any machine that exports it.

**Computed, not observed.** The helper says where tmux *would* put the socket, not that a daemon is listening. Its docstring says so and points live callers at the `#{socket_path}` format instead. `Server.__repr__` is the only non-test consumer, and it is display-only.

**Lives in `_internal`.** `libtmux._internal.env` already owns the tmux-environment readers (`resolve_env`, `socket_path_from_env`, `pane_id_from_env`), so this adds no public API surface.

**Empty `TMUX_TMPDIR` is treated as unset.** Verified against tmux 3.7b, where `TMUX_TMPDIR= tmux -L probe` also lands in `/tmp`.

## Scope

Repairing the unreachable `socket_path` auto-derivation in `Server.__init__` is deliberately **not** in this PR — that is a behaviour change with wider blast radius, tracked separately. This resolver would make it a small change, though.

## Test plan

- [x] `test_repr_socket_path_honors_tmux_tmpdir` — sets `TMUX_TMPDIR` via `monkeypatch` and asserts the repr names a path inside it; fails on `master`
- [x] `test_repr_socket_name` — covers the `getattr` removal (behaviour-preserving, so it passes either way)
- [x] Doctests on `Server.__repr__` and `resolve_socket_path`, which are collected and run
- [x] The three shipped doctests asserting the old output still pass unedited: `README.md`, `docs/quickstart.md`, `docs/api/libtmux.neo.md`
- [x] `just build-docs` clean

## Note for the merger

The `CHANGES` entry cites the issue number; swap it for this PR's number on merge. `CHANGES` will conflict with the sibling PRs for #724, #725 and #726, which all add to the same block — keep both sides.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,11 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Fixes
+
+- {class}`~libtmux.Server` now reprs the socket path tmux resolves from
+  `$TMUX_TMPDIR` instead of a hard-coded `/tmp/tmux-<euid>/default` (#723)
+
 ### Documentation
 
 #### Cleaner `from_env` examples (#719)
--- a/src/libtmux/_internal/env.py
+++ b/src/libtmux/_internal/env.py
@@ -1,4 +1,4 @@
-"""Readers for the tmux variables exported into every pane's environment.
+"""Readers for the tmux variables libtmux takes its bearings from.
 
 libtmux._internal.env
 ~~~~~~~~~~~~~~~~~~~~~
@@ -24,11 +24,16 @@ window is moved or linked into another session, the session id baked into
 libtmux therefore reads *only* the socket path out of ``TMUX`` and asks tmux
 itself -- targeting ``TMUX_PANE`` -- for the pane's window and session. See
 :meth:`libtmux.Pane.from_env`.
+
+A third variable, ``TMUX_TMPDIR``, is read *by* tmux rather than exported by
+it: it picks the directory tmux keeps its sockets in. See
+:func:`resolve_socket_path`.
 """
 
 from __future__ import annotations
 
 import os
+import pathlib
 import typing as t
 
 from libtmux import exc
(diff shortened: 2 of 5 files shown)


tony:
### Code review

Found 1 issue:

1. The `__repr__` doctest patches `TMUX_TMPDIR` with an unscoped `monkeypatch.setenv`, which leaks a live tmux daemon on every test run. The patch stays in effect through fixture finalization, so `_reap_test_server` resolves the socket under `/run/user/1000`, finds nothing, reports the server dead and skips `kill()` — then unlinks a path that does not exist.

https://github.com/tmux-python/libtmux/blob/8f437fab92e2ef3dc23cdc02bfc3211e4c135609/src/libtmux/server.py#L2698-L2702

Reproduced on this head:

```console
$ uv run pytest "src/libtmux/server.py::libtmux.server.Server.__repr__" --reruns 0 -q
1 passed
$ tmux -L libtmux_testch0bxj3k list-sessions
libtmux_35sr0x_t: 1 windows
tmuxp: 1 windows
```

A sibling doctest that does not `setenv` (`Server.is_alive`) leaves nothing behind, so the `setenv` is the cause rather than the doctest server itself.

`with monkeypatch.context() as m:` fixes it, and is the form the stacked #730 already uses for the same reason.

Issue #723, opened 2026-07-29
`Server.__repr__` hard-codes `/tmp/tmux-<euid>/default`, ignoring `$TMUX_TMPDIR`


tony:
*Filed against **tmux-python/libtmux** v0.62.0. Found while writing tested documentation examples: the repr appeared in a pytest failure banner and named a socket the object was not talking to.*

## What happens

`Server.__repr__` falls through to a hard-coded `/tmp/tmux-<euid>/default` when both `socket_name` and `socket_path` are `None`. tmux does not hard-code that path — it resolves the socket directory from `$TMUX_TMPDIR`, or `/tmp` when that is unset. `TMPDIR` is not consulted — `man tmux`: *"the sockets are all created in a directory tmux-UID under the directory given by TMUX_TMPDIR or in /tmp"*, confirmed empirically on tmux 3.7b (`TMPDIR=$t tmux` still lands in `/tmp/tmux-1000/default`). So under any `TMUX_TMPDIR` (test harnesses, sandboxes, containers, anyone with `TMUX_TMPDIR` exported in their shell profile) the repr names a socket the object is not using.

- [`src/libtmux/server.py#L2681-L2692`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/server.py#L2681-L2692) — the `__repr__` fall-through

```python
        return (
            f"{self.__class__.__name__}(socket_path=/tmp/tmux-{os.geteuid()}/default)"
        )
```

This is a *display* bug, not a behaviour bug — the object talks to the right socket. But repr is what shows up in tracebacks, pytest `--showlocals` banners, logs, and `print(server)`, which is exactly where someone is trying to work out which server they are looking at.

## Recreation

```python
import os, tempfile
os.environ["TMUX_TMPDIR"] = tempfile.mkdtemp(prefix="repro-")
from libtmux.server import Server

server = Server()
server.new_session(session_name="demo")
print("repr(server)      :", repr(server))
print("actual socket_path:", server.cmd("display-message", "-p", "#{socket_path}").stdout[0])
server.kill()
```

Observed on libtmux v0.62.0 / tmux 3.7b:

```
repr(server)      : Server(socket_path=/tmp/tmux-1000/default)
actual socket_path: /tmp/repro-pa15_hfr/tmux-1000/default
```

Expected: a repr that either names the real socket, or does not claim to know it.

## Why the fall-through is reachable at all

[...]

PR #728, opened 2026-07-29
Scoping a Server no longer kills it

tony:
> **Breaking change.** Read the Before / After below before merging.

## Summary

- **Fix** `with Server() as server:` destroying a tmux daemon it had nothing to do with. `__exit__` was `if self.is_alive(): self.kill()`, unconditionally — so a block that did no work at all took out the reader's default server and every session in it.
- **Change** the default: leaving the block kills nothing.
- **Add** `kill_on_exit` for when teardown *is* what you want.
- **Document** why a server differs from a session, window and pane, on `Server`, `__exit__`, and in `docs/topics/context_managers.md`.

## Why scoping is not ownership

A `Server` is a handle on a socket, not a connection. The same handle addresses a daemon whether or not this process started it, and a bare `Server()` addresses the default socket — usually the tmux the reader has been working in all day. Entering a `with` block says "I want this scoped", not "I am entitled to destroy this".

`Session`, `Window` and `Pane` are different in kind, and stay as they are: you got them by calling `new_session()` or `split()`, so the block really did create them.

## Before / After

```python
Server(socket_name="my-work").new_session(session_name="important")

with Server(socket_name="my-work") as server:
    pass

Server(socket_name="my-work").is_alive()   # False — the session is gone
```

```python
Server(socket_name="my-work").new_session(session_name="important")

with Server(socket_name="my-work") as server:
    pass

Server(socket_name="my-work").is_alive()   # True — nothing was asked for
```

Ask for teardown explicitly:

```python
with Server(socket_name="build", kill_on_exit=True) as server:
    server.new_session(session_name="ci")
```

## Design decisions

**Inference was implemented, then rejected.** An earlier revision of this branch had `__enter__` record whether the daemon was already alive and `__exit__` kill only one the block started. It reads well and has real precedent — `zipfile._filePassed`, `tarfile._extfileobj`, `gzip.myfileobj` all decide cleanup by whether they opened the resource. It was dropped for two reasons.

The first is a defect. The ownership signal was `is_alive()`, which catches bare `Exception` and returns `False` by contract — every way of failing to reach a server is a "no". So a transient tmux failure at `__enter__` would have read as "this block started it", and `__exit__` would have killed a daemon that predated the block. That is this issue's bug, reachable through a different door, and closing it would have meant hardening a deliberately lenient predicate into a strict one.

The second is that the stdlib's inference cases infer from *provenance* — what the object did — not from *liveness*. `Server` has no provenance to record: the socket does not say who started the daemon. Inference here was a guess wearing the costume of a fact.

**Explicit opt-in has its own precedent, including in languages built around ownership.** `subprocess.Popen.__exit__` unambiguously owns its child and still does not kill it. `sqlite3.Connection.__exit__` commits or rolls back and never closes. Rust's `std::process::Child` has no `impl Drop` at all — *"The standard library does not automatically wait on child processes (not even if the `Child` is dropped), it is up to the application developer to do so."*

**`kill_on_exit` is a plain `bool`, not a tri-state.** With no inference there is no third state to spell. It also matches the codebase: every one of libtmux's 188 `bool | None` parameters is consumed as `if x is not None and x:`, so `None` already means `False` here — a `None` meaning "maybe kill" would have inverted the house idiom in the destructive direction.

**The name is action-shaped on purpose.** libtmux already ships `Pane.display_popup(close_on_exit=...)`, and tmux's own vocabulary is `remain-on-exit`, `exit-empty`. An ownership-shaped name (`owns_server`) would assert a fact libtmux cannot verify.

**No `temp_server()` helper.** `temp_server` is already a doc anchor for the `TestServer` fixture, which creates auto-cleaned servers. A scoped-server helper can be designed on its own merits later.

## Test plan

- [x] `test_server_context_manager` — the default leaves a server alive, including one the block started. This previously asserted the opposite; the change is the breaking change made visible in the suite
- [x] `test_server_context_manager_spares_pre_existing_server` — a pre-existing server survives an empty block
- [x] `test_server_context_manager_spares_the_default_socket` — a **bare** `Server()` under a monkeypatched `$TMUX_TMPDIR`. Every other test routes through the `TestServer` factory, which mints a private socket and structurally cannot reproduce this issue's shape
- [x] `test_server_context_manager_kill_on_exit_true` — the opt-in kills a server the block did not start
- [x] `test_server_context_manager_kill_on_exit_true_when_block_started_it` — and one it did
- [x] Doctests on `Server`, `__enter__`, `__exit__`, plus the reworked `docs/topics/context_managers.md` — all collected and run
- [x] Full suite clean, including `libtmux.pytest_plugin`, which leans on server teardown
- [x] `just build-docs` clean

## Follow-up, not in this PR

`Session.from_session_id`, `Window.from_window_id` and `Pane.from_pane_id` all exist, so a child handle can be obtained without creating the object — `with Window.from_window_id(...)` kills a window you did not make. Same bug class, smaller blast radius, and worth its own issue rather than widening this one.

## Note for the merger

`CHANGES` will conflict with the sibling PRs for #723, #725 and #726, which all add to the same block — keep both sides. This one lands under `### Breaking changes`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,31 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Breaking changes
+
+#### Scoping a `Server` no longer kills it (#724)
+
+Leaving a `with` block used to kill any live {class}`~libtmux.Server`, so
+`with Server() as server:` around no work at all destroyed the default tmux
+daemon and every session in it. A server is a handle on a socket rather than a
+connection: the same handle addresses a daemon whether or not your process
+started it, so scoping one to a block is not grounds to destroy it.
+
+Teardown is now something you ask for, with the new
+{attr}`~libtmux.Server.kill_on_exit`. {class}`~libtmux.Session`,
+{class}`~libtmux.Window` and {class}`~libtmux.Pane` are unchanged — those you
+get by creating them. See {ref}`context_managers`.
+
+```python
+# Before: killed whatever daemon was on that socket
+with Server(socket_name="my-work") as server:
+    pass
+
+# After: nothing is killed unless you say so
+with Server(socket_name="my-work", kill_on_exit=True) as server:
+    pass
+```
+
 ### Documentation
 
 #### Cleaner `from_env` examples (#719)
--- a/docs/topics/context_managers.md
+++ b/docs/topics/context_managers.md
@@ -6,9 +6,12 @@ When you create tmux objects through libtmux, they normally live until you
 explicitly kill them. A context manager hands that cleanup back to Python: you
 scope an object to a block, and libtmux kills the underlying tmux object the
 moment you leave it — whether you exit cleanly or an exception unwinds the
-stack. The {class}`~libtmux.Server`, {class}`~libtmux.Session`,
-{class}`~libtmux.Window`, and {class}`~libtmux.Pane` classes (all main tmux
-objects) support this.
+stack. {class}`~libtmux.Session`, {class}`~libtmux.Window`, and
+{class}`~libtmux.Pane` work this way.
+
+{class}`~libtmux.Server` is the exception, and deliberately so: it does not
+kill itself on the way out unless you ask it to. A server is shared, and a
+handle on one says nothing about who started it.
 
 Most readers never reach for this. If you're building a long-running
 application, you typically let objects persist and tear them down yourself. The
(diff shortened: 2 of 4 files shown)


tony:
### Code review

Found 1 issue:

1. Two sections further down `context_managers.md` still promise the old behaviour, so the page now contradicts itself. The commit message says it reworked the page "which promised unconditional teardown for all four classes" — two of those promises survive.

https://github.com/tmux-python/libtmux/blob/7ea0444db340802e5444f5f250547bdd42594e9d/docs/topics/context_managers.md#L117-L141

"Nested context managers" still says the block gives you "every layer torn down for you", and item 4 reads "The server is killed when exiting its context". The `with Server() as server:` example immediately above it now leaves the daemon running.

https://github.com/tmux-python/libtmux/blob/7ea0444db340802e5444f5f250547bdd42594e9d/docs/topics/context_managers.md#L142-L149

"Benefits" then says "you never manually call the {meth}`~libtmux.Server.kill` ... methods", which is exactly what a reader now has to do.

The doctests cannot catch this: the nested block asserts no output, so the prose can drift from the behaviour silently. `docs/AGENTS.md` asks, in its pre-commit checklist, "Do the doctests run, and did you leave every code block, table, error string, and cross-reference exact?"

https://github.com/tmux-python/libtmux/blob/7ea0444db340802e5444f5f250547bdd42594e9d/docs/AGENTS.md#L145-L150

Issue #715, opened 2026-07-11
docs(pane_interaction, workspace_setup): the same shell-echo race #706 fixed in automation_patterns


tony:
#706 / #709 fix the echo race in {ref}`automation-patterns`. The identical pattern is still live on two other pages, and one of them is the page `automation_patterns.md` sends readers to *first*.

The fault is always the same: {meth}`~libtmux.Pane.send_keys` types the command into the pane, so the shell echoes it straight back into the buffer that `capture_pane()` reads. Any completion check that substring-scans the whole pane matches its own command line, and returns before the command has produced a single byte of output.

## `docs/topics/pane_interaction.md`

**`run_and_capture` ([v0.61.0](https://github.com/tmux-python/libtmux/blob/v0.61.0/docs/topics/pane_interaction.md))** is the worst instance, because it builds the marker into the command it sends:

```python
pane.send_keys(f'{command}; echo {marker}')     # the echo now contains __DONE__
...
output_str = '\n'.join(output)
if marker in output_str:                        # ...so this matches immediately
    return output
```

It returns on the first poll, in milliseconds, with only the echoed command on screen. Its `TimeoutError` branch is unreachable. The doc's own assertion (`'captured text' in '\n'.join(result)`) then passes off the echo line rather than the command's output, so the example looks like it works.

**`check_for_errors`** on the same page is the same helper #709 just fixed in `automation_patterns.md`, duplicated:

```python
output = '\n'.join(pane.capture_pane())
for pattern in error_patterns:
    if pattern in output:
        return True
```

A healthy `mytool --log-level=ERROR` is reported as a crash, because `ERROR` is in the command line the shell echoed.

## `docs/topics/workspace_setup.md`

A bare sleep followed by a whole-pane substring scan:

```python
time.sleep(0.2)
'Task A' in '\n'.join(pane_a.capture_pane())
```

Passes off the echo, not the output.

## Suggested fix

Same as #709, and it should be stated once and reused rather than re-derived per page:

- When the command is yours to change, synchronize with {meth}`~libtmux.Server.wait_for` instead of polling. `tmux wait-for` is race-free by construction: a signal with no waiter is remembered, so there is no lost wakeup.
- When a poll is genuinely needed, match **complete output lines that cannot also occur in the echoed input** — not substrings of the joined pane.
[...]

PR #729, opened 2026-07-29
Fix `retry_until`'s clock, and let it forward args

tony:
## Summary

- **Add** a keyword-only `args` tuple to `retry_until()`, unpacked into every call of the predicate, so the object being waited on is passed rather than closed over.
- **Add** two typing overloads: the zero-argument predicate keeps strict checking when `args` is absent, and `args` is required on the forwarding overload so it cannot be selected by accident.
- **Document** the parameter, with a doctest that waits on each pane of a loop.

## The problem

A zero-argument predicate has nowhere to put the subject, so a wait written inside a loop has to close over the loop variable. This repo enables flake8-bugbear, which rejects that:

```
B023 Function definition does not bind loop variable `pane`
    retry_until(lambda: any(pane.capture_pane()))
                             ^^^^
```

B023's documented fix is a default-argument binding — which mypy then refuses to infer:

```
error: Cannot infer type of lambda  [misc]
```

So the lint-clean form does not type-check and the type-clean form does not lint, leaving `# type: ignore[misc]` as the only way out at every such call site.

**No existing call site on `master` suffers from this** — they all use named nested functions. This is a preventative ergonomic change, not a repair of live breakage.

## Before / After

```python
for pane in panes:
    retry_until(lambda pane=pane: any(pane.capture_pane()))  # type: ignore[misc]
```

```python
for pane in panes:
    retry_until(lambda p: any(p.capture_pane()), args=(pane,))
```

## Design decisions

**Overloads rather than one widened signature.** Widening `fun` to `Callable[..., bool]` outright would have stopped mypy catching a predicate that wrongly takes arguments in the common zero-argument case. Two overloads keep that: the first takes `Callable[[], bool]` and has no `args` parameter at all, the second takes `Callable[..., bool]` with `args` **required** — no default — so the strict overload is selected whenever `args` is absent.

**`args` is keyword-only.** `seconds` is already positional-second, so a positional `args` would have been ambiguous at the call site and a source of silent misreads.

**Not shipping ready-made waits.** The issue also proposes exposing the two common cases ("this pane produced anything", "this pane printed this whole line") as methods. That is a larger, separate change with its own API surface, and would deserve its own issue; this PR only removes the reason a caller has to write a closure.

## A trap the examples deliberately avoid

`capture_pane()` returns the command tmux echoed onto the pane, so a substring test for the expected output matches that echo and is already true before the shell has run anything. Both the new doctest and the new test therefore compare **whole lines**, and say why in a comment — otherwise they would pass in a world where no shell ever ran, which is exactly the failure mode being tracked in #715.

## Test plan

- [x] `test_retry_until_forwards_args_from_loop` — waits on each pane of a loop through the new path, with no `# type: ignore` and no `# noqa` in the file
- [x] `test_retry_until_args_reach_predicate` — the tuple arrives at the predicate on every attempt
- [x] `uv run ruff check .` clean, which is the assertion that B023 no longer fires
- [x] `uv run mypy` clean, which is the assertion that the overloads resolve
- [x] Doctests on `retry_until` collected and run
- [x] `just build-docs` clean

## Note for the merger

The `CHANGES` entry cites the issue number; swap it for this PR's number on merge. It will conflict with the sibling PRs for #723, #724 and #725, which all add to the same block — keep both sides.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,16 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### Fixes
+
+- {func}`~libtmux.test.retry.retry_until` measures its budget with
+  {func}`time.monotonic` instead of the wall clock, so an NTP correction, a
+  container clock sync, or a VM resume can no longer cut a wait short or
+  stretch it (#726)
+- {func}`~libtmux.test.retry.retry_until` accepts an `args` tuple forwarded to
+  the predicate, so waiting on a loop variable no longer needs a closure that
+  ruff's `B023` rejects and mypy cannot infer (#726)
+
 ### Documentation
 
 #### Cleaner `from_env` examples (#719)
--- a/src/libtmux/test/retry.py
+++ b/src/libtmux/test/retry.py
@@ -70,6 +70,13 @@ def retry_until(
     raises : bool
         Whether or not to raise an exception on timeout. Defaults to ``True``.
 
+    Notes
+    -----
+    The budget is measured with :func:`time.monotonic`, which only ever moves
+    forward. A wall clock does not: an NTP correction, a container clock sync,
+    or a VM resume steps it, and a step landing inside the wait would end it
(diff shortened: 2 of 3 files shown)


tony:
### Code review

Issue #660, opened 2026-04-19
pytest plugin leaks tmux daemons and socket files (libtmux_test* servers not reaped)


tony:
## Summary

The pytest plugin shipped with `libtmux` creates per-test tmux servers on `libtmux_test*` sockets via its `server` / `session` / etc. fixtures, but teardown does not reliably kill the tmux daemons or unlink the socket files. Over repeated runs this accumulates both live processes and stale socket files under `/tmp/tmux-<uid>/`. Downstream consumers of the fixtures (e.g. `tmux-python/libtmux-mcp`) inherit the leak.

Surfaced by `libtmux-mcp`'s new `list_servers` MCP tool (0.1.0a2), which lists **43 live** `libtmux_test*` daemons on a dev machine alongside the user's real `default` server. On disk the same machine shows **10,090** `libtmux_test*` socket files under `/tmp/tmux-1000/`.

Reported downstream at tmux-python/libtmux-mcp#20; filing here as the likely root-cause location.

## Evidence

From a dev machine after normal development activity:

```
$ # via libtmux-mcp's list_servers (0.1.0a2) — 43 live libtmux_test* servers
$ ls /tmp/tmux-1000 | grep -c libtmux_test
10090
$ pgrep -af 'tmux.*libtmux_test' | wc -l
# non-zero — varies, but > 0 after any recent test run
```

The disparity (≈43 live, ≈10 k on disk) suggests two distinct leaks:

- **Live leak:** some fixtures exit without `server.kill_server()`, so the daemon stays up indefinitely.
- **Socket-file leak:** when daemons do exit, the socket file is not `unlink`ed, so `/tmp/tmux-<uid>/` accumulates inodes forever.

## Reproduction

1. Clone `tmux-python/libtmux` (or any downstream that uses its pytest plugin, e.g. `tmux-python/libtmux-mcp`). `uv sync`.
2. `N0=$(ls /tmp/tmux-$(id -u) | grep -c libtmux_test)` — baseline.
3. `uv run pytest`.
4. `N1=$(ls /tmp/tmux-$(id -u) | grep -c libtmux_test)` — post-run count. `N1 > N0`.
5. `pgrep -af 'tmux.*libtmux_test'` — observe one or more live tmux daemons on `libtmux_test*` sockets.

## Expected behavior

After `pytest` exits cleanly (pass or fail), **zero new** `libtmux_test*` processes remain and **zero new** socket files remain under `/tmp/tmux-<uid>/`.

[...]

PR #730, opened 2026-07-30
Measure a tmux socket path where tmux resolves it

### LINKED ISSUE #725: `Server` does not validate socket path length; overrun surfaces as a raw tmux `File name too long`

tony:
Fixes #725. Stacked on #727.

A tmux socket is a UNIX domain socket, so its path is capped by `sockaddr_un` — 107 bytes on Linux, 103 on macOS. Over that, tmux says:

```
error connecting to /tmp/dddd…/tmux-1000/default (File name too long)
```

That names the path, but not how far over the limit it is, nor which variable made it that long. This adds `exc.SocketPathTooLong` carrying all three.

## Where the measurement happens

The routes are not equivalent, so they are not measured in the same place.

`socket_path=` is handed to tmux unchanged as `-S<path>`. Nothing can revise it, so it is measured at construction, where the caller still holds the frame that wrote it.

`socket_name=` is not. libtmux passes `-L<name>` and tmux resolves the directory from `$TMUX_TMPDIR` **when it runs**. Measured on tmux 3.7b: construct a `Server(socket_name=…)` with `TMUX_TMPDIR=a`, change it to `b`, then dispatch — tmux binds under `b`. So a construction-time check on that route tests a value with no bearing on what tmux will use. It passes when the variable later grows (hitting the exact `File name too long` this exists to prevent) and refuses when it later shrinks. Those paths are measured per command, at the same point `colors` is already validated.

A bare `Server()` inside a tmux pane is left alone entirely. A bare tmux client prefers `$TMUX` over `$TMUX_TMPDIR`, so a deep socket directory is irrelevant to it — measured: with `$TMUX` set, `tmux list-sessions` succeeds while `$TMUX_TMPDIR` names a directory far too deep to bind. Checking it would have refused a server tmux reaches without difficulty, in the case libtmux is most often used from.

## `Server.socket_args()`

The `-S`/`-L` flags were being reconstructed in four places — `Server.cmd()`, `Server.raise_if_dead()`, the format-query layer, and the control-mode client, the last two spawning tmux through their own `Popen`/`check_call`. A check added to any one of them would have covered part of the dispatch surface. They now all build their argv from `Server.socket_args()`, so the measurement is total by construction rather than by discipline.

## Constructing a server has no side effect

Naming a server is not using one. A named or bare `Server` always builds, so `Server.is_alive()` keeps answering for a server that cannot be reached — an address the kernel cannot hold is one more way of not being alive — and `Server.raise_if_dead()` keeps being the way to ask why. `Server.sessions` stays lenient and returns an empty `QueryList`, matching the documented contract for list accessors.

## Verification

`ruff check`, `ruff format`, `mypy` (69 files), `py.test --reruns 0` (1477 passed, 1 skipped), `just build-docs` — all clean. Green on tmux 3.2a through master.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,10 +45,42 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### What's new
+
+#### Socket paths are measured before tmux sees them (#730)
+
+A tmux socket is a UNIX domain socket, so its path is capped by `sockaddr_un`
+— 107 bytes on Linux, 103 on macOS. {class}`~libtmux.Server` now measures the
+path and raises the new {exc}`~libtmux.exc.SocketPathTooLong`, carrying the
+byte count, how far over the limit it is, and where the length came from.
+tmux reports the overrun as `error connecting to <path> (File name too long)`,
+which names the path but not the numbers, nor which variable made it long.
+
+Where the measurement happens follows what tmux actually reads. A
+`socket_path` is passed through unchanged as `-S<path>`, so it is measured at
+construction, where the caller can still change it. A `socket_name` resolves
+against `$TMUX_TMPDIR` — which tmux re-reads when it runs, not when the
+{class}`~libtmux.Server` was built — so it is measured on each command
+instead. The inherited case is the one that bites: a pytest `tmp_path`, an XDG
+runtime dir, a nested worktree, a CI checkout under a long workspace prefix.
+The fix is a shorter socket directory: {func}`tempfile.mkdtemp` or a short
+`$TMUX_TMPDIR`. See {ref}`socket_path_length` for the pytest case.
+
+Naming a server that way stays free of side effects, so
+{meth}`~libtmux.Server.is_alive` keeps answering — an address the kernel
+cannot hold is one more way of not being alive — and
+{meth}`~libtmux.Server.raise_if_dead` keeps being the way to ask why.
+
+A bare {class}`~libtmux.Server` inside a tmux pane is left alone. tmux prefers
+`$TMUX` over `$TMUX_TMPDIR` when no socket is named, so a script running
+inside tmux is measured against the socket it will actually use rather than a
+directory tmux never consults.
+
 ### Fixes
 
-- {class}`~libtmux.Server` now reprs the socket path tmux resolves from
-  `$TMUX_TMPDIR` instead of a hard-coded `/tmp/tmux-<euid>/default` (#723)
+- {class}`~libtmux.Server` now reprs the socket a bare tmux client would use —
+  `$TMUX` inside a pane, otherwise the path resolved from `$TMUX_TMPDIR` —
+  instead of a hard-coded `/tmp/tmux-<euid>/default` (#727)
 
 ### Documentation
 
--- a/docs/api/testing/pytest-plugin/usage.md
+++ b/docs/api/testing/pytest-plugin/usage.md
@@ -112,6 +112,58 @@ True
 
 This is particularly useful when testing interactions between multiple tmux servers or when you need to verify behavior across server restarts.
 
+(socket_path_length)=
+
+### Socket paths and the UNIX socket limit
+
+A tmux socket is a UNIX domain socket, so its path is capped by `sockaddr_un`
+— 107 bytes on Linux, 103 on macOS. pytest's {fixture}`tmp_path` is nested
+deep by design (`/tmp/pytest-of-<user>/pytest-<n>/<test-name><n>`), so putting
+a socket under it — directly, or by pointing `TMUX_TMPDIR` at it — can overrun
+the limit on a long test name or a long temporary root. {class}`~libtmux.Server`
+measures an explicit `socket_path` as soon as it is passed and raises
+{exc}`~libtmux.exc.SocketPathTooLong` with the byte count, rather than letting
+tmux report `File name too long` with only the path to go on:
+
+```python
+>>> from libtmux import exc
+>>> from libtmux.server import Server as TmuxServer
+>>> deep_socket = "/tmp/" + "d" * 120 + "/sock"
+>>> try:
+...     TmuxServer(socket_path=deep_socket)
+... except exc.SocketPathTooLong as e:
+...     print(e.length)
+130
+```
+
+A `socket_name` is different: tmux resolves it against `$TMUX_TMPDIR` when it
+runs, so the length is only knowable at dispatch. Building the object is safe,
+and a test that only asks whether a server is there gets an answer instead of an
+exception — an unbindable address is one more way of not being alive:
+
+The directory has to exist for that to be the socket tmux would bind: tmux takes
+the first of `$TMUX_TMPDIR` and `/tmp` that resolves.
+
+```python
+>>> from libtmux.server import Server as TmuxServer
+>>> deep = request.getfixturevalue("tmp_path") / ("d" * 120)
+>>> deep.mkdir()
+
+>>> with monkeypatch.context() as m:
+...     m.delenv("TMUX", raising=False)
+...     m.setenv("TMUX_TMPDIR", str(deep))
+...     TmuxServer(socket_name="deep").is_alive()
+False
+```
+
+The fixtures in this plugin sidestep it: {fixture}`server
+<libtmux.pytest_plugin.server>` and {fixture}`TestServer
+<libtmux.pytest_plugin.TestServer>` name their sockets with `socket_name`, which
+tmux resolves under its own short socket directory. In your own tests, keep
+`tmp_path` for files and reach for {func}`tempfile.mkdtemp` — which gives a
+short `/tmp/<random>` — when you need a socket path of your own, or point
+`TMUX_TMPDIR` somewhere short.
+
 (set_home)=
 
 ### Setting a temporary home directory
--- a/docs/topics/configuration.md
+++ b/docs/topics/configuration.md
@@ -46,13 +46,16 @@ without you arranging anything.
 
 That leaves the two variables that *are* yours to set, and most people
 set neither. `TMUX_TMPDIR` is tmux's own — the directory it keeps sockets
-in. libtmux never reads it, but the tmux binary it shells out to does, so
-it shapes which server a bare {class}`~libtmux.Server` lands on; pass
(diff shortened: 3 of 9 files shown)


tony:
### Code review

Found 1 issue:

1. The `socket_name` branch validates a socket path libtmux *computes*, but tmux resolves symlinks before binding, so a configuration tmux handles fine is now refused at construction with no opt-out.

https://github.com/tmux-python/libtmux/blob/f3771f77a4124a29d613f6872f1d1aeea39e44e6/src/libtmux/server.py#L227-L240

Reproduced on this head, with `TMUX_TMPDIR` set to a 93-char symlink pointing at a short directory:

```console
$ TMUX_TMPDIR=$LONG tmux -f /dev/null display-message -p '#{socket_path}'
/tmp/sd3394337/tmux-1000/default
```

tmux `realpath()`s the socket directory in `make_label()`, so the real path is 34 bytes and binds without complaint. libtmux's `resolve_socket_path()` does not, so it measures the unresolved 93-char symlink instead:

```
SocketPathTooLong: Socket path for socket_name='probe' is 109 bytes, over the 107 byte limit
```

`decee4bf` removed this same computation from `Server.__init__` and gave the reason in its body — "Guessing the path would have been wrong anyway. tmux resolves a -L name against its own socket directory, which honours TMUX_TMPDIR, the per-uid directory and the platform's overrides. Callers that need the real path ask tmux for it" — and pinned the contract with `test_socket_path_not_derived_from_socket_name`. #727 reintroduced the guess but kept it harmless: repr text and a best-effort unlink, with a docstring repeating the caveat that the path is "*computed*, not observed". This PR turns the same guess into a hard failure.

The `socket_path is not None` branch is sound, since tmux uses `-S` verbatim with no `realpath()`. Only the resolved-from-`socket_name` branch inherits the guess. Options: `os.path.realpath()` the socket directory before measuring, or keep the eager check for the caller-supplied `socket_path` and defer the resolved-name case to where the path meets tmux.

tony:
Force-pushed a redesign of where the check runs. The earlier revision validated the socket path in `Server.__init__` on every route; two measurements against tmux 3.7b showed that is unsound for the routes that resolve through the environment.

**tmux reads `$TMUX_TMPDIR` at exec, not at construction.** Constructing `Server(socket_name=N)` with `TMUX_TMPDIR=a`, changing it to `b`, then dispatching, binds under `b`. A construction-time check on that route therefore tests a value with no bearing on what tmux uses — it passes when the variable later grows, which is the exact `File name too long` this issue is about, and refuses when it later shrinks.

**A bare client prefers `$TMUX` over `$TMUX_TMPDIR`.** With `$TMUX` set, `tmux list-sessions` succeeds while `$TMUX_TMPDIR` names a directory far too deep to bind. The previous revision refused to construct a bare `Server()` there — in the case libtmux is most often used from, a script running inside tmux.

`socket_path=` is unaffected and is still refused at construction, since libtmux passes it through unchanged as `-S<path>`.

This also closes the design question raised in review: `Server.is_alive()` stays total, because a named or bare `Server` always constructs.

tony:
### Code review

Found 1 issue, now fixed in 337c9211:

1. `test_bare_server_inside_a_pane_ignores_tmux_tmpdir` patched `$TMUX_TMPDIR` for the whole test, leaking a live tmux daemon on every run. The `server` fixture takes `monkeypatch`, so pytest undoes the patches only after that fixture's finalizer has run — `_reap_test_server` then resolves the socket under the 120-char directory, cannot reach the server so skips `kill()`, and unlinks a path that never existed. Both halves of the reap contract fail silently while the test passes.

https://github.com/tmux-python/libtmux/blob/593d5ffd2ba4a1a0e9ac0e1fa06ba4a6e40b93cd/tests/test_server.py#L323-L332

Reproduced deterministically on that revision: the test passes, then `tmux -L libtmux_test<...> list-sessions` lists two live sessions. This is the same failure mode raised on the parent PR in this stack (#727), and it defeats the invariant `_reap_test_server` was written for (#660, stale sockets accumulating indefinitely).

Fixed by scoping the patches to the assertions with `monkeypatch.context()`, matching how #727 resolved it. Verified: no sockets left behind after the test, and none after the full `tests/test_server.py` run.

Also corrected in the same push: `Server.socket_args()`'s docstring said the check "belongs at dispatch rather than at construction", which stopped being accurate once an explicit `socket_path` became a construction-time check. It now states the split.

tony:
### Code review

Found 3 issues, all fixed in 66ff46f9:

1. libtmux refused servers tmux reaches. tmux creates `<dir>/tmux-<euid>` and silently binds under `/tmp` when it cannot, so a `$TMUX_TMPDIR` that does not exist never holds the socket however long its name is. Measuring it anyway made `is_alive()` report a live daemon dead and the first command raise.

https://github.com/tmux-python/libtmux/blob/c8356b9d0e5e3f9b4e9c9e2f0e4f3d2a1b6c7e8f/src/libtmux/_internal/env.py#L143-L149

Measured: `TMUX_TMPDIR=/tmp/<200 d's> tmux -L probe new-session -d` exits 0 with the socket at `/tmp/tmux-1000/probe`, while libtmux reported `is_alive() -> False` and raised `SocketPathTooLong`. `resolve_socket_path()` now mirrors the fallback; a directory that exists and is too deep still raises.

2. `__repr__` named a socket the object does not use. It resolved from `$TMUX_TMPDIR` even inside a pane, where a bare client uses `$TMUX` — the rule this PR itself establishes for `socket_args()`.

3. Three tests and a doctest pinned the false refusal by pointing `$TMUX_TMPDIR` at a nonexistent directory. They now use a real one, so they still exercise the true positive.

tony:
### Code review

Found 2 issues, fixed in 57d931d0:

1. The previous round's fix was only half applied. `Server.is_alive`'s doctest and the pytest-plugin usage page still pointed `$TMUX_TMPDIR` at a directory that does not exist, so after the `/tmp` fallback landed they stopped reaching the length check at all — they passed for the wrong reason.

https://github.com/tmux-python/libtmux/blob/4ad7f9c7e6b3c9a1f0d2e8b7a4c5d6e9f0a1b2c3/src/libtmux/server.py#L367-L372

Both now create the directory first, so the resolved path is one tmux would really bind.

2. The `Server` class docstring built its directory with `tempfile.mkdtemp()`, leaking one per run. It now uses the `tmp_path` fixture through `request`, which pytest rotates.

tony:
### Code review

Found 2 issues, fixed in adcc295f:

1. The stated tmux mechanism was wrong in four places. "tmux creates `<dir>/tmux-<euid>` and quietly falls back to `/tmp` when it cannot" describes the uncreatable case, which is a hard error rather than a fallback. Measured on tmux 3.7b: a nonexistent directory and a broken symlink both fall back and exit 0, while a directory that exists but is unwritable, or is a regular file, both fail with `couldn't create directory`. tmux takes the first of `$TMUX_TMPDIR` and `/tmp` that *resolves*.

https://github.com/tmux-python/libtmux/blob/57d931d0e0c9b0a0f8d6e2c1a3b5d7f9e0a2c4b6/src/libtmux/_internal/env.py#L111-L116

2. `is_dir()` was the wrong predicate for that rule. `exists()` follows symlinks the way `realpath()` does, so a broken link falls back as tmux does, and a regular file is left to tmux to reject rather than silently resolving to a `/tmp` path tmux would never bind. Verified both the original regression and the true positive still behave correctly.

PR #733, opened 2026-07-31
Bound the control-mode wait off the descriptor

### LINKED ISSUE #731: `test_control_mode_stdout_preserves_non_ascii_output` polls `select` on a buffered stream
*Filed against **tmux-python/libtmux** v0.62.0, tmux 3.7b. Flaked twice in CI on unrelated pull requests, on different tmux matrix cells, both times passing on re-run.*

## What happens

`test_control_mode_stdout_preserves_non_ascii_output` bounds its wait with `select.select()` on a **buffered** stream and then consumes with `readline()`, so readiness and readability are measured against two different buffers.

[`tests/test_control_mode.py#L88-L94`](https://github.com/tmux-python/libtmux/blob/v0.62.0/tests/test_control_mode.py#L88-L94)

`ControlMode` builds its subprocess with `text=True, encoding="utf-8"`, so `stdout` is a `TextIOWrapper` over a `BufferedReader`:

[`src/libtmux/_internal/control_mode.py#L83-L90`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/_internal/control_mode.py#L83-L90) and [`#L99`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/_internal/control_mode.py#L99)

`select` calls `fileno()` and sees only the kernel pipe. One `readline()` drains up to 8192 bytes off that pipe into userspace and returns the first line; the rest of the reply sits in the decoder where `select` cannot see it, and the descriptor is now empty.

tmux frames each command's control-mode output with `%begin` / `%end` guards written in one burst, so the whole reply arrives together:

[`cmd-queue.c#L825-L832`](https://github.com/tmux/tmux/blob/3.7b/cmd-queue.c#L825-L832)

Observed timeline from a raw `os.read` on the same descriptor:

```
t=0.0002  read(68)   %begin ... \n%end ... \n%session-changed $0 s\n
t=0.0003  read(50)   %begin ... \n\xe2\x90\x9e\n%end ... \n
t=0.4619  read(137)  %output %0 ...
```

The separator is fully buffered by the first or second `readline()`. What normally rescues the test is that third read: unrelated `%output` from the pane's shell painting its prompt re-arms `select`, and the loop then serves its remaining lines out of the userspace buffer. **The test passes for a reason unrelated to the data it asserts on.**

## Reproduction

Deterministic once the read loop starts after the prompt-paint burst has landed, which is what a loaded CI runner produces. Injecting a stall between `ControlMode.__enter__` and the loop, with everything else verbatim:

```
delay=0.0   0/4 failed
delay=0.15  1/4 failed
delay=0.3   3/4 failed
```

Proof the data was in hand at the moment `select` reported not-ready:

```
select NOT ready at iter=1 (last read = '%begin 1785458429 355 1')
  readline #0 returned in 7us  -> '␞'
  readline #1 returned in 3us  -> '%end 1785458429 355 1'
  readline #2 BLOCKED (fd empty)
```

`select` burned its full one-second budget while the answer was seven microseconds away.

## CI signature

```
tests/test_control_mode.py:91: in test_control_mode_stdout_preserves_non_ascii_output
    assert ready, "timed out waiting for control-mode output"
E   AssertionError: timed out waiting for control-mode output
E   assert []
        line       = '%begin 1785370231 290 0\n'
        ready      = []
```

`line` is the loop variable holding the last line successfully read, so in both failures the test read `%begin` and the next `select` timed out. That is the signature of "read one line, then the descriptor went quiet", not of a slow tmux.

## Ruled out, with evidence

- **The `locale.setlocale(LC_CTYPE, "C")` the test performs.** Removing it changes nothing: 4/4 failures at `delay=0.3`.
- **tmux version.** The mechanism is entirely Python-side. Two flakes landing on two of eight matrix cells is what randomness looks like.
- **`bufsize` on `Popen`.** Buffering mode is a write-side setting; measured no effect for `0`, `1`, or default.

## What a fix needs

The library code is correct and should not change. `text=True, encoding="utf-8"` is deliberate — it is the regression this test exists to guard.

1. **Drop `select`; read on a thread with a deadline.** A reader thread pushes lines onto a `queue.Queue`, the test polls the queue with a timeout. Nothing straddles two buffers. Verified that a threaded reader still catches the original regression: with `encoding="utf-8"` removed it raises `UnicodeDecodeError`, so the coverage is preserved.
2. **Or `select` on the raw descriptor** via `ctl.stdout.buffer.raw` and decode in the test. Correct, but the test then owns partial-line handling and decoding.

Option 1 is preferred: the only reason `select` is present is to bound the wait, and a thread with a deadline does that without introducing a second buffer.

## Related

`tmux wait-for` is the general answer for authored commands and is already wrapped as `Server.wait_for()` ([`src/libtmux/server.py#L617`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/server.py#L617), [`cmd-wait-for.c#L34-L38`](https://github.com/tmux/tmux/blob/3.7b/cmd-wait-for.c#L34-L38)), but it does not apply here: this test waits on control-mode protocol output rather than on a shell command it authored.

tony:
Closes #731.

## Summary

- **Fix** `test_control_mode_stdout_preserves_non_ascii_output` timing out while the line it wants sits in the buffer. It bounded its wait with `select` on the descriptor and read with `readline`, which serves from the `TextIOWrapper` above it.
- **Replace** `select` with a reader thread and a queue polled against a monotonic deadline, so the wait is bounded without asking the descriptor a question it cannot answer.
- **Leave `ControlMode` untouched.** `text=True, encoding="utf-8"` is exactly the behaviour this test exists to guard.

## Why the descriptor cannot answer

tmux writes a whole `%begin`/`%end` block in one burst, so the reply arrives together. The first `readline` drains up to 8192 bytes off the descriptor into userspace and returns one line; the rest sits in the decoder, and the descriptor is now empty. `select` reports not-ready while the answer is microseconds away:

```
select NOT ready at iter=1 (last read = '%begin 1785458429 355 1')
  readline #0 returned in 7us  -> '␞'
  readline #1 returned in 3us  -> '%end 1785458429 355 1'
  readline #2 BLOCKED (fd empty)
```

What made the old test pass was unrelated `%output` from the pane's shell painting its prompt, arriving ~460 ms later and re-arming `select`. The test passed for a reason unconnected to the data it asserts on, which is why it failed on loaded CI runners: once the read loop starts after that burst, failure is certain rather than unlikely.

## Before / After

Injecting a 300 ms stall before the read loop, everything else verbatim:

```
before:  delay=0.0  0/4 failed   delay=0.15  1/4 failed   delay=0.3  3/4 failed
after:   delay=0.3  0/4 failed
```

## The coverage is preserved

The risk in rewriting this test is silently retiring what it guards. Verified by removing `encoding="utf-8"` from `ControlMode` and re-running:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 107: ordinal not in range(128)
1 failed
```

Still caught. Restored, it passes.

## Design notes

**A thread rather than `select` on `stdout.buffer.raw`.** Both work. Selecting on the raw descriptor means the test owns partial-line assembly and decoding — real logic, in a test whose subject is decoding. The thread keeps `readline` doing the line splitting and adds only a bound.

**Monotonic deadline, not a per-iteration timeout.** A fixed timeout per line lets a slow-but-progressing stream run for `limit × timeout`. The deadline bounds the whole read.

**Fails, never hangs.** Stream exhaustion is reported distinctly from expiry, so a closed pipe does not present as a timeout.

## Test plan

- [x] `tests/test_control_mode.py` green, 3 consecutive runs at `--reruns 0`
- [x] Survives a 300 ms injected stall, 4/4, where the previous version failed 3/4
- [x] Still fails with `encoding="utf-8"` removed from `ControlMode`
- [x] `just build-docs` clean

## Note for the merger

`CHANGES` will conflict with the sibling PRs for #723, #724, #725 and #726, which all add to the same block — keep both sides.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -59,6 +59,13 @@ it.
 
 ### Development
 
+#### The control-mode test no longer flakes in CI (#733)
+
+`test_control_mode_stdout_preserves_non_ascii_output` intermittently timed out
+waiting for output that had already arrived, so it failed on loaded runners and
+passed on re-run. The test now bounds its wait with a reader thread and a
+deadline. The non-ASCII decoding regression it guards is still covered.
+
 #### CI actions updated to current majors
 
 Workflow actions moved to their current major releases: `actions/checkout` v7,
--- a/tests/test_control_mode.py
+++ b/tests/test_control_mode.py
@@ -4,8 +4,10 @@ from __future__ import annotations
 
 import locale
 import os
-import select
+import queue
 import sys
+import threading
+import time
 import typing as t
 
 import pytest
@@ -14,7 +16,63 @@ from libtmux._internal.control_mode import ControlMode
 from libtmux.formats import FORMAT_SEPARATOR
 
 if t.TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from libtmux.server import Server
+
+
+def _read_lines(
+    stream: t.IO[str],
+    *,
+    limit: int,
+    timeout: float,
+) -> Iterator[str]:
+    """Yield up to *limit* lines from *stream*, giving up after *timeout*.
+
+    A reader thread owns the stream and the caller polls a queue, so the wait
+    is bounded without anything having to ask the file descriptor whether a
+    line is available.
+
+    That question has no useful answer here. ``ControlMode`` builds its
+    subprocess with ``text=True``, so ``stream`` is a ``TextIOWrapper`` over a
+    ``BufferedReader``: ``select`` would report readiness on the raw
+    descriptor while ``readline`` serves from the userspace buffer above it.
+    tmux writes a whole ``%begin``/``%end`` block in one burst, so the first
+    ``readline`` routinely drains every remaining line off the descriptor --
+    leaving ``select`` with nothing to report and the answer already in hand.
+    """
+    lines: queue.Queue[str | BaseException | None] = queue.Queue()
+
+    def pump() -> None:
+        try:
+            for line in stream:
+                lines.put(line)
+        except BaseException as e:  # noqa: BLE001
+            # Carry it across the thread boundary. Collapsing it into the
+            # ``None`` sentinel would report the reader as having reached EOF,
+            # naming the wrong cause for a decode error this test exists to
+            # catch.
+            lines.put(e)
+        finally:
+            lines.put(None)
+
+    reader = threading.Thread(target=pump, daemon=True)
+    reader.start()
+
+    deadline = time.monotonic() + timeout
+    for _ in range(limit):
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            pytest.fail("timed out waiting for control-mode output")
+        try:
+            line = lines.get(timeout=remaining)
+        except queue.Empty:
+            pytest.fail("timed out waiting for control-mode output")
+        if isinstance(line, BaseException):
+            raise line
+        if line is None:
+            pytest.fail("control-mode stream closed before the expected output")
+        yield line
 
 
 def test_control_mode_creates_client(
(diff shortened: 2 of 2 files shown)


tony:
### Code review

tony:
### Code review

Issue #732, opened 2026-07-31
`Server.wait_for()` has no timeout, so a pane that never signals hangs forever


tony:
*Filed against **tmux-python/libtmux** v0.62.0, tmux 3.7b.*

## What happens

[`Server.wait_for()`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/server.py#L617) blocks until the channel is signalled, with no way to bound the wait. If the pane that was going to signal dies first — the command was killed, the shell exited, the window was closed — the call blocks forever.

There is no ceiling anywhere below it either. `tmux_cmd` calls `communicate()` with no timeout:

[`src/libtmux/common.py#L339`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/common.py#L339)

So the hang is process-wide and not interruptible except by signal.

`wait-for` blocking indefinitely is correct tmux behaviour, not a tmux bug — the command is a rendezvous and has no timeout flag:

[`cmd-wait-for.c#L34-L38`](https://github.com/tmux/tmux/blob/3.7b/cmd-wait-for.c#L34-L38)

```
.args = { "LSU", 1, 1, NULL },
.usage = "[-L|-S|-U] channel",
```

The gap is that libtmux exposes the rendezvous without exposing a way to give up on it.

## Why it matters now

`wait-for` is the exact, poll-free way to wait for a command a caller authored: the command signals a channel when it finishes and the caller blocks on that channel. Measured on tmux 3.7b — a pane running `sleep 1.5; tmux wait-for -S <ch>` released the waiter after 1.506 s, with zero polling and no round trips. A signal sent *before* the waiter starts is remembered, so there is no lost-wakeup race: signalling first and then waiting returned in 3 ms.

That makes it the right thing to teach in documentation for "run a command, then read its output", replacing the `time.sleep()` guesses currently in the topic pages. The blocker is that every such example either carries `subprocess` timeout ceremony or teaches a hang.

Concretely, the shape wanted in a doc example is:

```python
pane.send_keys(f"make; tmux wait-for -S {channel}")
server.wait_for(channel, timeout=60)
```

and today the second line cannot be written.

## Reproduction

A pane that never signals, because the keys were typed but never entered:

```python
pane.send_keys(f"echo READY; tmux wait-for -S {channel}", enter=False)
server.wait_for(channel)   # blocks forever
```

The bound is achievable one level down, which is why this is a missing parameter rather than a missing capability:

```python
[...]

PR #734, opened 2026-07-31
Automation patterns: wait for tmux instead of guessing

tony:
## Summary

- **Fix** a doctest on {ref}`automation-patterns` that fails deterministically today at `--reruns 0`, hidden by the repo's `--reruns=2` default.
- **Replace** fixed sleeps with a tmux channel the command signals when it finishes, blocked on with `Server.wait_for()` — which returns the moment the command is done rather than after a guess.
- **Replace** substring searches of the joined capture with whole-line comparisons, because `capture_pane()` returns the command tmux echoed onto the pane.
- **No new API.** Uses `send_keys`, `capture_pane` and `Server.wait_for`, all of which ship today.

## Why the sleeps could not work

Measured on this branch: `send_keys` to visible in `capture_pane()` is **p50 100.7 ms** under `pytest -n auto`, p90 130 ms. The page used `time.sleep(0.1)`. That is roughly **-1 ms of headroom** at the median.

## The bug this fixes is live

`docs/topics/automation_patterns.md` on `master`, four consecutive runs at `--reruns 0`:

```
1 failed, 15 passed
1 failed, 15 passed
1 failed, 15 passed
1 failed, 15 passed
```

Same command on this branch:

```
16 passed
16 passed
16 passed
16 passed
```

The failing example is "Capturing output between markers", whose helper searches `'\n'.join(pane.capture_pane())` for a marker the command itself contains — so it matches the echoed command line and returns before the command has produced anything. Its `TimeoutError` branch is unreachable. The page's own assertion then passes off the echo.

## The two changes

**Signal a channel, block on it.** tmux remembers a signal sent before the waiter starts, so there is no lost-wakeup race regardless of ordering. Measured: a pane running `sleep 1.5; tmux wait-for -S <ch>` released the waiter after 1.506 s, with no polling; signalling first and then waiting returned in 3 ms.

**Compare whole lines.** `capture_pane()` already returns `list[str]`; joining it and searching for a substring is what lets the echo satisfy the check. Demonstrated with `enter=False`, so no shell ever runs:

```
capture: ['$ echo READY']
'READY' in '\n'.join(capture)  -> True     <- the echo
'READY' in capture             -> False
```

Sleeps remain only where an example is about elapsed time itself.

## Test plan

- [x] `docs/topics/automation_patterns.md` green 4/4 at `--reruns 0`, against 0/4 on `master`
- [x] `just build-docs` clean, no new warnings

## Note for the merger

The `CHANGES` heading cites `(#NN)` — swap it for this PR's number on merge. It will conflict with the sibling PRs for #723, #724, #725, #726 and #731, which all add to the same block; keep both sides.

Related but not required: `Server.wait_for()` still has no timeout (#732), so a command that dies before signalling blocks forever. That does not affect the examples here, all of which signal on every exit path, but it is worth closing before this pattern is recommended more widely.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,20 @@ _Notes on the upcoming release will go here._
 
 ### Documentation
 
+#### Automation patterns waits for tmux instead of guessing (#734)
+
+The examples on {ref}`automation-patterns` waited for a command to finish by
+sleeping a fixed amount and then reading the pane. A sleep long enough to be
+safe is dead time on every run, and one short enough to feel quick reports
+failure on a busy machine.
+
+They now have the command signal a channel when it finishes and block on that
+with {meth}`Server.wait_for() <libtmux.Server.wait_for>`, which returns the
+moment the command is done. Checks compare whole captured lines rather than
+searching the joined capture, because {meth}`~libtmux.Pane.capture_pane`
+returns the command tmux echoed onto the pane -- a substring search matches
+that echo and is already true before the shell has run.
+
 #### Cleaner `from_env` examples (#719)
 
 The rendered examples for {meth}`Pane.from_env() <libtmux.Pane.from_env>` and
--- a/docs/topics/automation_patterns.md
+++ b/docs/topics/automation_patterns.md
@@ -12,18 +12,19 @@ makes that coordination ordinary Python — you start commands with
 pile of {meth}`send_keys() <libtmux.Pane.send_keys>` calls into automation you
 can trust: output monitoring, timeouts, retries, and multi-pane orchestration.
 
-Most scripts only need a couple of these patterns. Output monitoring and the
-context manager patterns cover the common case — send a command, wait for a
-marker, clean up after yourself — so start there. The later sections (state
+Most scripts only need a couple of these patterns. Command completion and the
+context manager patterns cover the common case — send a command, wait for it to
+finish, clean up after yourself — so start there. The later sections (state
 machines, task queues) are for the rarer cases where a single pane drives a longer
 sequence of steps; reach for them when you actually need them.
 
-These patterns lean on polling: you call {meth}`~libtmux.Pane.capture_pane` in a
-loop and `sleep` between reads. That is simpler than wiring up an event-driven
-system, and it costs you latency — each poll is a tmux round-trip, and a `sleep`
-between polls is dead time you pay whether the command finished or not. For most
-automation that trade is worth it. When milliseconds matter, look instead at tmux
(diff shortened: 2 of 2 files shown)


tony:
### Code review

Found 1 issue, now fixed in d4c8f7fe:

1. The new deliverable was inserted with its own `### Documentation` header instead of being appended under the existing one, so the unreleased entry carried two `### Documentation` sections and the `from_env`/ruff/tmux-options entries ended up under a detached second block. `CLAUDE.md` says "Fixed subheadings, in this order when present: `### Breaking changes`, `### Dependencies`, `### What's new`, `### Fixes`, `### Documentation`, `### Development`" — one of each.

https://github.com/tmux-python/libtmux/blob/a09ab07df1054f90685ccd0f2485e31655900933/CHANGES#L62-L67

The same hunk shipped the literal placeholder `(#NN)` in the deliverable heading; that is now `(#734)`.

tony:
### Code review

Found 1 issue, fixed in 3f81cf82:

1. The "Detecting errors in output" section claimed the tagged scan tells the two commands apart. It does not — `command_output()` reads the whole pane, so the check after the second command also sees the first command's `out:ok`. The example passed by accumulation rather than by discrimination, and a later success would still have reported an earlier failure.

https://github.com/tmux-python/libtmux/blob/80c8797c1b0f8b4e2f6a0c9d3e5a7b1c4d8e9f0a/docs/topics/automation_patterns.md#L166-L170

The prose now says what the tag actually does — it separates the command's output from the shell's echo, not one command's output from the last one's — and the failing command clears the pane first, so `command_output()` returns `['Error: no disk']` on its own.

tony:
### Code review

Found 1 issue, fixed in 5d304cb4:

1. Four polling examples used a 2.0 s budget, which does not cover cold shell startup when the suite runs in parallel. CI invokes `py.test ... -n auto`, so these were being carried by the `--reruns=2` in `addopts` rather than passing outright. Raised to 10.0 s; the page then passes 3/3 under `-n auto`, against 3 failures before.

https://github.com/tmux-python/libtmux/blob/3f81cf82a9d4e5b6c7f8a0b1c2d3e4f5a6b7c8d9/docs/topics/automation_patterns.md#L337-L339

Worth noting for the record: under a full-suite `-n auto` run on a many-core machine, this page also fails on `master` with identical counts, so the residual flakiness there is pre-existing and not introduced here. The budget change is an improvement rather than a regression fix — a timeout is a ceiling, not a cost.

tony:
### Code review

PR #735, opened 2026-07-31
Bound a tmux command that may never return

### LINKED ISSUE #732: `Server.wait_for()` has no timeout, so a pane that never signals hangs forever

tony:
Closes #732. Stacked on #734, which owns `docs/topics/automation_patterns.md`.

## What this does

`Server.wait_for()` wraps `tmux wait-for`, a rendezvous with no clock on it. When whatever was going to signal the channel dies first — the command is killed, the shell exits, the window closes — the wait never ends, and there was no ceiling anywhere beneath it either: [`communicate()`](https://github.com/tmux-python/libtmux/blob/v0.62.0/src/libtmux/common.py#L339) was called without one.

```python
pane.send_keys(f"make; tmux wait-for -S {channel}")
server.wait_for(channel, timeout=60)
```

`timeout` is keyword-only and available on `wait_for()` and on `cmd()` for `Server`, `Session`, `Window` and `Pane`, forwarding to `tmux_cmd`. Every entry point defaults to `None`, which is unbounded, so no existing call changes behaviour.

## Why the bound rides on the call

An instance-level default (`Server(timeout=30)`) was built and measured before this was chosen. It works, and it breaks blocking-by-design commands: on `Server(timeout=1.0)`, `server.cmd("attach-session", ...)` is killed at ~1.06s. Mitigating it needs a hand-maintained deny-list that by construction cannot cover a command a user dispatches themselves.

The reason is that the analogy to HTTP clients does not hold. `httpx` gets away with a client-level timeout because requests are homogeneous; tmux's command set is not. `display-message` should take milliseconds and `attach-session` should take hours, so no single per-server number is correct, and it is wrong precisely where being wrong hurts most.

Per-call also stays well-defined under batching, which matters for the engine work: a timeout carried by the request has an unambiguous meaning in a batch, where an engine-level default or an ambient scope has to choose between bounding the batch and bounding each command in it.

## Killing is not enough

`communicate()` does not kill the child when its timeout expires, and `kill()` alone leaves a zombie. Both are covered, because the difference is invisible to the obvious assertion — `os.kill(pid, 0)` succeeds against a zombie. The test pins `returncode == -SIGKILL` and that `os.waitpid` raises `ChildProcessError`.

`TmuxCommandTimeout` carries the killed command line and the bound it exceeded. It subclasses `WaitTimeout`, so handlers that already catch libtmux's timeout keep working, and it deliberately does not subclass `subprocess.TimeoutExpired` — the stdlib type never crosses the library boundary, including by chaining.

## Two tmux behaviours documented

Both were measured against tmux 3.7b while building this, and both are properties of the rendezvous rather than of this change.

A wait that times out is abandoned rather than withdrawn. tmux only remembers a signal when nothing is waiting for the channel ([`cmd-wait-for.c#L148`](https://github.com/tmux/tmux/blob/3.7b/cmd-wait-for.c#L148)), so the next signal is spent waking the abandoned wait — one signal, after which the channel behaves normally. A fresh channel name per rendezvous avoids it.

A returning wait is also weaker evidence than it looks. tmux releases every waiter when the server shuts down ([`cmd-wait-for.c#L244`](https://github.com/tmux/tmux/blob/3.7b/cmd-wait-for.c#L244)), and does so exactly as if the channel had been signalled, so a pane that died and a command that finished are indistinguishable from the wait alone. When the outcome matters, have the command report its own exit status.

## Verification

`ruff check`, `ruff format`, `mypy` (69 files), `py.test --reruns 0` (1468 passed, 1 skipped) and `just build-docs` all clean.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,31 @@ $ uvx --from 'libtmux' --prerelease allow python
 _Notes on the upcoming release will go here._
 <!-- END PLACEHOLDER - ADD NEW CHANGELOG ENTRIES BELOW THIS LINE -->
 
+### What's new
+
+#### Bounded tmux commands (#735)
+
+{meth}`Server.wait_for() <libtmux.Server.wait_for>` takes a `timeout` in seconds.
+A tmux channel is a rendezvous with no clock on it: when the pane that was going
+to signal is killed, exits early, or has its window closed, the waiter blocks for
+the life of the process. `server.wait_for(channel, timeout=60)` turns that into
+a {exc}`~libtmux.exc.TmuxCommandTimeout` you can catch. It subclasses
+{exc}`~libtmux.exc.WaitTimeout`, so existing handlers keep working, and it
+carries the command that was killed and the bound it exceeded.
+
+The bound is available to any tmux command, not just `wait-for`.
+{meth}`Server.cmd() <libtmux.Server.cmd>`,
+{meth}`Session.cmd() <libtmux.Session.cmd>`,
+{meth}`Window.cmd() <libtmux.Window.cmd>`,
+{meth}`Pane.cmd() <libtmux.Pane.cmd>`, and
+{class}`~libtmux.common.tmux_cmd` beneath them all accept `timeout`, and all
+kill and reap the tmux client libtmux spawned before raising, so the call
+leaves no child of its own behind. Work the command started — a pane's
+foreground process, the tmux server — keeps running. The default, `None`,
+waits as long as tmux takes.
+
+See {ref}`automation-patterns` for waiting on a channel instead of polling.
+
 ### Documentation
 
 #### Automation patterns waits for tmux instead of guessing (#734)
--- a/docs/topics/automation_patterns.md
+++ b/docs/topics/automation_patterns.md
@@ -126,8 +126,9 @@ block on that channel with {meth}`~libtmux.Server.wait_for`. tmux remembers a si
 sent before the waiter starts, so this has no lost-wakeup race. It also avoids
 confusing the shell's echoed command with the command's output.
 
-{meth}`~libtmux.Server.wait_for` has no timeout. Make sure every expected exit path
-reaches `tmux wait-for -S` so a failed command cannot leave your script blocked.
+{meth}`~libtmux.Server.wait_for` waits indefinitely unless you give it a `timeout`.
+Either make sure every expected exit path reaches `tmux wait-for -S`, or bound the
+wait, so a failed command cannot leave your script blocked.
 Channels are server-wide, so give each in-flight command a distinct channel name.
 
 ```python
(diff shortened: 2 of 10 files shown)


tony:
### Code review

tony:
### Code review

Found 1 issue, fixed in d172f7ca:

1. "no orphan survives" overstated what is reaped. The kill covers the tmux client libtmux spawned; the pane's foreground process and the tmux server keep running, so a caller reading that could believe a timeout stopped the work.

https://github.com/tmux-python/libtmux/blob/39e87af1d0a4c8b2e5f7a9c1b3d6e8f0a2c4d5e6/src/libtmux/exc.py#L361-L365

Both `TmuxCommandTimeout` and `tmux_cmd`'s `Raises` section now say the call leaves no child of its own behind, and that work the command started is unaffected.

tony:
### Code review

tony:
### Code review

Found 1 issue, fixed in 9378bede:

1. The previous round's correction to what a timeout actually kills landed in `exc.py` and `common.py` but not in the four public `cmd()` docstrings or `CHANGES`, which still said the timeout kills "tmux" and that "an expiry leaves nothing behind".

https://github.com/tmux-python/libtmux/blob/e4e1f920d3b8a5c7f9e1d2a4b6c8e0f2a4c6d8e0/CHANGES#L64-L68

All five now say the same thing: the kill covers the tmux client libtmux spawned, and work the command started keeps running.

